PR #7, opened 2025-11-21
Project Import: Use allowed namespaces when importing

sniok:
Requires #8 

---

Use allowed namespaces when importing
Cancel namespace websocket connection after checking if the project is aks one

### DIFF
--- a/plugins/aks-desktop/src/components/ImportAKSProjects/ImportAKSProjects.tsx
+++ b/plugins/aks-desktop/src/components/ImportAKSProjects/ImportAKSProjects.tsx
@@ -260,6 +260,14 @@ export default function ImportAKSProjects() {
           continue;
         }
 
+        // add allowed namespaces
+        const settings = JSON.parse(
+          localStorage.getItem(`cluster_settings.${clusterName}`) || '{}'
+        );
+        settings.allowedNamespaces ??= [];
+        settings.allowedNamespaces.push(...namespacesInCluster.map(it => it.namespace.name));
+        localStorage.setItem(`cluster_settings.${clusterName}`, JSON.stringify(settings));
+
         // Step 2: Mark all namespaces in this cluster as successfully imported
         // No need to patch labels - they already exist from AKS managed namespace
         for (const { namespace, projectName } of namespacesInCluster) {


plugins/aks-desktop/src/components/ImportAKSProjects/ImportAKSProjects.tsx
```diff
@@ -260,6 +260,14 @@ export default function ImportAKSProjects() {
           continue;
         }
 
+        // add allowed namespaces
+        const settings = JSON.parse(
+          localStorage.getItem(`cluster_settings.${clusterName}`) || '{}'
+        );
```
joaquimrocha, line 266: We have `getAllowedNamespaces` which we export to plugins AFAIU. We shouldn't use the actual settings' keys for preventing their eventual change in the future.

joaquimrocha (changes requested):

plugins/aks-desktop/src/components/ImportAKSProjects/ImportAKSProjects.tsx
sniok: yeah but there's no exposed method to set allowed namespaces. I think we need to treat localStorage keys as public api anyway since some plugins are already using it directly

illume: I agree we should use APIs for this... But IMHO we can delay this to later as long as an issue is made to fix it later.

illume: @sniok mark this as resolved? (We can leave making the API in headlamp for later?)

plugins/aks-desktop/src/components/ImportAKSProjects/ImportAKSProjects.tsx
```diff
@@ -260,6 +260,14 @@ export default function ImportAKSProjects() {
...
+        settings.allowedNamespaces ??= [];
+        settings.allowedNamespaces.push(...namespacesInCluster.map(it => it.namespace.name));
```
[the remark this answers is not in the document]
illume: I guess this can't hurt doing.

plugins/aks-desktop/src/components/ImportAKSProjects/ImportAKSProjects.tsx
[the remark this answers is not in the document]
illume: Same here. It's a valid point IMHO.

PR #8, opened 2025-11-21
upstreamable: headlamp: Improve namespace list handling when allowedNamespaces is set. Small fixes.

sniok:
This PR adds a special handling for a case when:
 - Namespace.useList is used
 - allowedNamespaces are set for the cluster

In this case instead of requesting /namespaces list it will make individual requests for namespace and create a list from those items. This is done because when allowedNamespaces are set user typically doesn't have `list` access for namespaces.

---

Other small fixes: don't retry queries when response has status codes like 403, fix view button activity location

### DIFF
--- a/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
+++ b/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
@@ -17,6 +17,7 @@
 import type { QueryObserverOptions } from '@tanstack/react-query';
 import { useQueries, useQueryClient } from '@tanstack/react-query';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { getAllowedNamespaces } from '../../cluster';
 import type { KubeObject, KubeObjectClass } from '../../KubeObject';
 import type { QueryParameters } from '../v1/queryParameters';
 import { ApiError } from './ApiError';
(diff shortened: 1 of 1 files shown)


frontend/src/lib/queryClient.ts
```diff
@@ -21,6 +21,19 @@ export const queryClient = new QueryClient({
...
       staleTime: 3 * 60_000,
       refetchOnWindowFocus: false,
+      retry(failureCount, error: unknown) {
+        // Do not retry for unauthorized or bad requests
+        if (
+          !!error &&
+          typeof error === 'object' &&
+          'status' in error &&
+          !isNaN(Number(error.status)) &&
+          Number(error.status) < 500
```
joaquimrocha, line 31: What if it's a temporary 404?

frontend/src/components/resourceMap/sources/definitions/relations.tsx
```diff
@@ -167,7 +167,7 @@ const endpointSlicesToServices = makeRelation(
 );
 
 const ingressToService = makeRelation(Ingress, Service, (ingress, service) =>
-  ingress.spec.rules?.find((rule: any) =>
+  ingress.spec?.rules?.find((rule: any) =>
```
joaquimrocha: Seems unrelated to the PR?

frontend/src/lib/queryClient.ts
sniok: I mean you could ask that question for pretty much any status code but I think we should have some sensible defaults where we don't have a spinner for long time for requests that most likely won't succeed

frontend/src/components/resourceMap/sources/definitions/relations.tsx
sniok: I'll move it to another PR

PR #20, opened 2025-12-03
upstreamable: frontend: Map: add more defensive checks in relations

sniok:
Adds more null checks for networking objects to avoid crashes.


illume (changes requested):
I guess this should be done in Headlamp repo (or maybe you have already?), then we bring it in with the next Headlamp sync?

The CI is failing, moving to draft.

PR #71, opened 2025-12-29
github: Improve workflow security

skoeva:
These changes update the GitHub Actions workflows by pinning an action by hash and setting top-level permissions where they aren't defined, improving security.

### DIFF
--- a/.github/workflows/build-app-linux.yml
+++ b/.github/workflows/build-app-linux.yml
@@ -11,6 +11,9 @@ on:
         required: true
         default: 'main'
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-app-mac.yml
+++ b/.github/workflows/build-app-mac.yml
@@ -11,6 +11,9 @@ on:
         required: true
         default: 'main'
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: macos-latest
--- a/.github/workflows/build-app-win.yml
+++ b/.github/workflows/build-app-win.yml
@@ -11,6 +11,9 @@ on:
         required: true
         default: 'main'
 
+permissions:
+  contents: read
+
 jobs:
   build:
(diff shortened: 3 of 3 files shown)


skoeva:
@microsoft-github-policy-service agree company="Microsoft"

.github/workflows/build-app-linux.yml
illume, line 40: Upload I think needs some other permission?

.github/workflows/update-headlamp-submodule.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout main repository with submodules
-        uses: actions/checkout@v6
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```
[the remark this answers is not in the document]
skoeva: good catch, this should have been updated

.github/workflows/build-app-linux.yml
illume: Sorry, I was wrong about this.

illume (approved):
🎉 thanks

Tatsinnit (approved):

.github/workflows/build-app-linux.yml
Tatsinnit, line 1: 💡  Just some thoughts, shall we fix this as well: `no topLevel permission defined` https://github.com/Azure/aks-desktop/security/code-scanning/6 (we can always do it separately), I noticed to gently FYI, thanks.

illume:
> please: both commits are not verified aka signed-off by: - so DCO check should fail or we should add that, and the commit sign-off should be a requirement

There is a CLA check, so DCO is not necessary.

.github/workflows/build-app-linux.yml
skoeva: This is already covered in the commit:
```yml
permissions:
     contents: read
```

skoeva (commented):
closing in favor of #77

Issue #38, opened 2025-12-12
plugins/aks-desktop: Untangle imports, get rid of barrel files, add CI check


illume:
- There are some circular loops of imports we need to get rid of
- Import types properly (so don't import whole module when just need a type)
- Remove index barrel files for components to simplify the codebase and make it faster to load and easier to edit
- Add a CI check to make sure we don't add more index files

PR #93, opened 2026-01-05
plugins/aks-desktop: Clean up imports and barrel files

skoeva:
These changes do the following:
- Remove circular imports
- Remove full module imports when type imports suffice
- Remove unnecessary barrel files (these are bad)
- Add an eslint rule to detect barrel files in the code

Fixes: #38 

### Testing
- [x] `cd plugins/aks-desktop && npm run build` does not fail


illume (dismissed):
thanks! 🎉

plugins/aks-desktop/scripts/check-no-barrel-files.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
sniok: let's use an eslint rule for this instead of custom script 

https://www.npmjs.com/package/eslint-plugin-no-barrel-files

plugins/aks-desktop/src/utils/azure/index.ts
```diff
@@ -8,6 +8,9 @@
  * Azure CLI wrapper, resource client factory, and Graph API integration.
  */
 
+// Azure type definitions
```
sniok: this is a barrel file

.github/workflows/ci.yml
```diff
@@ -71,6 +71,9 @@ jobs:
       - name: Check Prettier formatting
         run: npm run format -- --check
 
+      - name: Check for barrel index files
+        run: npm run lint:no-barrels
```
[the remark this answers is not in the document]
skoeva: removed this step because it's no longer necessary

plugins/aks-desktop/scripts/check-no-barrel-files.sh
skoeva: done

sniok (approved):

Issue #29, opened 2025-12-10
[BUG] Unable to run on Ubuntu 24.04.3


pauldotyu:
## Bug Description

Downloaded the latest from the releases page and installed using the `dpkg -i  aks-desktop_0.1.0-alpha-1_amd64.deb`. When attempting to run the application, it fails to launch.

## Expected Behavior

The application should launch without errors

## Actual Behavior

The application is installed into the `/opt/AKS desktop` directory and when attempting to launch the app from the terminal, I see the following error message:

```
$ /opt/AKS\ desktop/aks-desktop
LaunchProcess: failed to execvp:
/opt/AKS
[336292:1210/135905.387465:FATAL:zygote_host_impl_linux.cc(201)] Check failed: . : Invalid argument (22)
[1]    336292 trace trap (core dumped)  /opt/AKS\ desktop/aks-desktop
```

## Environment

- **AKS desktop version**: 0.1.0-alpha-1
- **OS**: Ubuntu 24.04.3 LTS x86_64

## Additional Context

It appears the space in the folder name `/opt/AKS desktop` is causing issues when the app is launching. Log seems to indicate that it is attempting to launch the app from the `/opt/AKS` directory which is non-existent.

Here's my workaround:

```
# rename the folder
sudo mv /opt/AKS\ desktop /opt/AKS-desktop

# recreate the symlink 
sudo ln -sfn /opt/AKS-desktop/aks-desktop /etc/alternatives/aks-desktop

# update desktop launcher
sudo sed -i 's|/opt/AKS desktop|/opt/AKS-desktop|g' /usr/share/applications/aks-desktop.desktop

# optionally refresh desktop database
sudo update-desktop-database /usr/share/applications
```

PR #235, opened 2026-02-10
aksd: fix: Override productName for Linux builds

tejhan:
## Description

Overrode productName for Linux builds to resolve error where builds couldn't be executed on some linux environments.

The `productName` "AKS desktop" caused electron-builder to install the app to `/opt/AKS desktop/`. In some environments,  Chromium's zygote process, which manages sandboxed child processes on Linux, did not properly handle spaces in the executable path — it splits `/opt/AKS desktop/aks-desktop` into `/opt/AKS` and `desktop/aks-desktop`, causing the launch to fail.

This issue does not appear to affect Windows or macOS, which use different process spawning mechanisms that handle the spaces correctly.

## Changes Made

Override `productName` to `AKS-Desktop` (hyphenated) in `electron-builder.config.js` when building exclusively for Linux. This
changes the install path to `/opt/AKS-Desktop/`, avoiding the space.

The override is guarded so it only applies when `--linux` is passed without `--win` or `--mac`, ensuring Windows and macOS builds retain the original "AKS desktop" branding.

## Related Issues

- https://github.com/Azure/aks-desktop/issues/29
Closes #[29]

## Testing

1. Built the deb package with and without the fix
2. Reproduced the original crash on a native Ubuntu 24.04 VM with multipass
3. Verified the fixed build installs to `/opt/AKS-Desktop/` and launched successfully

### DIFF
--- a/app/electron-builder.config.js
+++ b/app/electron-builder.config.js
@@ -64,6 +64,15 @@ if (config.deb) {
   };
 }
 
+// Override productName for Linux builds to avoid spaces in installation path
+if (
+  process.argv.includes('--linux') &&
+  !process.argv.includes('--win') &&
+  !process.argv.includes('--mac')
+) {
+  config.productName = 'AKS-Desktop';
+}
+
 // On non-Mac platforms, ensure dmg-license is truly optional by not including Mac targets during --dir builds
 // This prevents electron-builder from trying to validate Mac-specific dependencies on Linux/Windows
 if (process.env.npm_lifecycle_event === 'build' && process.platform !== 'darwin') {


app/electron-builder.config.js
```diff
@@ -64,6 +64,15 @@ if (config.deb) {
   };
 }
 
+// Override productName for Linux builds to avoid spaces in installation path
+if (
+  process.argv.includes('--linux') &&
+  !process.argv.includes('--win') &&
+  !process.argv.includes('--mac')
+) {
```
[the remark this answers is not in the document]
tejhan: Yes this caveat was considered- i intentionally excluded --win & --mac for this reason actually. If we only go off the --linux flag, then we'll end up renaming the windows & mac distributions as well for the multi-OS build. 

Went for less renaming rather than more.

If we intend to rename the productName for all builds maybe we could eliminate this entirely, but that's up to discussion.

app/electron-builder.config.js
```diff
@@ -64,6 +64,15 @@ if (config.deb) {
   };
 }
 
+// Override productName for Linux builds to avoid spaces in installation path
```
sniok, line 67: let's instead define `executableName` in package.json as `aks-desktop` for all platforms, I think that'll be simpler

https://www.electron.build/configuration#executablename

sniok: this will allow us to keep the "AKS Desktop" with a space as a product name

tejhan: I tried this as an initial solution, but upon testing, it appeared that defining `executableName` didn't change the actual installation path. `productName` seems to have highest precedence there. 

So the app would end up still installing to /opt/AKS desktop/aks-desktop, which did still yield the initial bug observed, where we get a Chromium Zygote error still when attempting to run the binary on that path due to the space still being there. (On native ubuntu)

Will double check to verify. I'm using an Ubuntu VM to replicate the customer scenario using dkpg & then running out of the /opt/ directory.

sniok: that makes sense, you're right

PR #244, opened 2026-02-12
refactor: extract runCommandAsync to shared module

gambtho:
## Summary
- Extract `runCommandAsync()` into `utils/shared/run-command.ts` — a promise-based wrapper around Headlamp's `pluginRunCommand` bridge, so it can be reused by other integrations
- Simplify `az-cli.ts` to delegate to the shared module (no signature changes — callers are unaffected)
- Fix stderr handling when logging out of Azure CLI

## Test plan
- [x] `run-command` unit tests pass (9 tests)
- [x] TypeScript type check passes
- [x] ESLint + Prettier pass

### DIFF
--- a/plugins/aks-desktop/src/components/AzureAuth/AzureProfilePage.tsx
+++ b/plugins/aks-desktop/src/components/AzureAuth/AzureProfilePage.tsx
@@ -35,7 +35,13 @@ export default function AzureProfilePage() {
     try {
       // Import dynamically to avoid circular dependencies
       const { runCommandAsync } = await import('../../utils/azure/az-cli');
-      await runCommandAsync('az', ['logout']);
+      const result = await runCommandAsync('az', ['logout']);
+
+      if (result.stderr && result.stderr.includes('ERROR:')) {
+        console.error('Azure CLI logout error:', result.stderr);
+        setLoggingOut(false);
+        return;
+      }
 
       // Trigger update event for sidebar label
       window.dispatchEvent(new CustomEvent('azure-auth-update'));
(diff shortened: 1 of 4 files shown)


plugins/aks-desktop/src/components/AzureAuth/AzureProfilePage.tsx
```diff
@@ -34,8 +34,14 @@ export default function AzureProfilePage() {
     setLoggingOut(true);
     try {
       // Import dynamically to avoid circular dependencies
-      const { runCommandAsync } = await import('../../utils/azure/az-cli');
-      await runCommandAsync('az', ['logout']);
+      const { tryExec } = await import('../../utils/azure/az-cli');
+      const result = await tryExec('az', ['logout']);
+
+      if (result.stderr) {
```
sniok: this commit contains two changes, refactoring run command functions and handling stderr when logging out, could you split those in two commits please?

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -29,66 +27,20 @@ function needsRelogin(error: string): boolean {
...
-  });
+function runAzCommandAsync(args: string[]): Promise<{ stdout: string; stderr: string }> {
+  const actualCommand = getAzCommand();
+  debugLog('[AZ-CLI] Command resolution: az →', actualCommand);
+  debugLog('[AZ-CLI] Executing command:', actualCommand, 'with args:', args);
+  return runCommandAsync(actualCommand, args);
 }
 
-async function tryExec(
+export async function tryExec(
```
sniok: I think we don't need this function tbh since it's just calling `runCommandAsync`. We can just use `runCommandAsync` and move the `az` command resolution inside it. But this is just a nit feel free to ignore

sniok (dismissed):
thanks, I've left a couple of comments

plugins/aks-desktop/src/utils/azure/az-cli.ts
gambtho: good call, i split this up from a  large PR, but reviewing in context it doesn't provide any benefit

Issue #258, opened 2026-02-17
New Project Compute Quota Tab: Fix heading element for "Memory Resources"


vyncent-t:
Related to issue #202 

Description:
When following the Create Project workflow in AKS desktop, clicking the "+ Create Project" button at the main tab, then **clicking "AKS Managed Project"**, then continue into "Compute Quota" tab-

_**Under the "Compute Quota" tab, notice that the  "Memory Resources" header is not in sequentially-descending order**_

Lighthouse scan flagged as the following:

"Heading elements are not in a sequentially-descending order"

"Properly ordered headings that do not skip levels convey the semantic structure of the page, making it easier to navigate and understand when using assistive technologies."

How to test: 

- Navigate to the Projects tab
- Click the "+ Create Projects" button
- Click "AKS Managed Project" option
- Continue progress into "Compute Quota" tab
- Use inspect tools and open Lighthouse
- Scan at this view
- Note that the same logic used for "CPU Resources" is used for "Memory Resources"

Issue #202, opened 2026-01-28
a11y - Create AKS project (Compute Quota Tab - Feature Flag Required heading element)


vyncent-t:
This issue an an a11y issue related to workflow Create AKS Project https://github.com/Azure/aks-desktop/issues/194

Description:
When following the Create Project workflow in AKS desktop, clicking the "+ Create Project" button at the main tab, then **clicking "AKS Managed Project"**, then continue into "Compute Quota" tab-

_**Under the "Compute Quota" tab, notice that the "Feature Flag Required" header is not in sequentially-descending order**_

Lighthouse scan flagged as the following:

"Heading elements are not in a sequentially-descending order"

"Properly ordered headings that do not skip levels convey the semantic structure of the page, making it easier to navigate and understand when using assistive technologies."

How to test: 

- Navigate to the Projects tab
- Click the "+ Create Projects" button
- Click "AKS Managed Project" option
- Continue progress into "Compute Quota" tab
- Use inspect tools and open Lighthouse
- Scan at this view

---

Lighthouse:

<img width="875" height="613" alt="Image" src="https://github.com/user-attachments/assets/e9dacd9a-d718-4270-9ed2-d1c272eaeea6" />

View:

<img width="1812" height="997" alt="Image" src="https://github.com/user-attachments/assets/0c20ddcb-b5be-43fb-980e-f4a1286612fd" />

<img width="926" height="808" alt="Image" src="https://github.com/user-attachments/assets/df1996a3-13fa-4a59-8cdc-1ed2a1c55c7c" />

Issue #219, opened 2026-02-06
a11y: fix bugs found (umbrella issue)


illume:
Working on these now will de-risk things for making GA deadline.

Estimate is added here rather than in the 11 sub issues. @vyncent-t the 2 week estimate is for all of the sub issues.
Aggressive but possible: 1 day, Likely certain it will be done: 2 weeks

This issue does not cover the issues the accessibility team finds.

PR #259, opened 2026-02-17
frontend: Fix CPU resources & Memory Resources headers

vyncent-t:
## Description

This PR fixes _**a collection**_ of a11y issues identified by Lighthouse where headings were not in sequential order for the  Create AKS Project workflow "Compute Quota" tab.

The ResourceCard component is used for sub categories "CPU Resources" and "Memory Resource" on the "Compute Quota" tab of the Create AKS Project workflow. Both of the sub categories lack a proper semantic header order. This PR introduces a fix for both of these issues by adding a component prop to the titles.

# Related Issues

- Closes https://github.com/Azure/aks-desktop/issues/258
- Closes https://github.com/Azure/aks-desktop/issues/202

Related to #219 

## Changes Made

- Fixes Lighthouse scans flagged as the following:
"Heading elements are not in a sequentially-descending order"

"Properly ordered headings that do not skip levels convey the semantic structure of the page, making it easier to navigate and understand when using assistive technologies."

How to test: 

- Navigate to the Projects tab
- Click the "+ Create Projects" button
- Click "AKS Managed Project" option
- Continue progress into "Compute Quota" tab
- Use inspect tools and open Lighthouse
- Scan at this view

## Images

image

image

### DIFF
--- a/plugins/aks-desktop/src/components/CreateAKSProject/components/ResourceCard.tsx
+++ b/plugins/aks-desktop/src/components/CreateAKSProject/components/ResourceCard.tsx
@@ -14,7 +14,7 @@ export const ResourceCard: React.FC<ResourceCardProps> = ({ title, icon, iconCol
     <Box>
       <Box sx={{ display: 'flex', alignItems: 'center', mb: 2 }}>
         <Icon icon={icon} width={24} height={24} color={iconColor} style={{ marginRight: 12 }} />
-        <Typography variant="h6" sx={{ fontWeight: 'bold' }}>
+        <Typography variant="h6" component="h3" sx={{ fontWeight: 'bold' }}>
           {title}
         </Typography>
       </Box>


vyncent-t:
> see [#257 (review)](https://github.com/Azure/aks-desktop/pull/257#pullrequestreview-3815142581)
> 
> I also think for continued work on fixing these headers, they could all be committed in the same PR since they are all small & related changes

As per the audit effort, each of the issues sourced in #219 are individual wrongs that need correcting. They also require an organized paper trail, complete with sufficient testing instructions and a direct resolution.

Plus, most of these issues are rooted in their own workflow, so the "how to test" section for a combined PR would be massive.

I have already grouped together fixes that share a common workflow / view ( https://github.com/Azure/aks-desktop/pull/246 and https://github.com/Azure/aks-desktop/pull/259) to reduce review workload

sniok (approved):

PR #260, opened 2026-02-17
aksd: Add secure storage API and GitHub OAuth browser flow

gambtho:
## Summary
- Add secure storage IPC handlers via Electron's `safeStorage` API (DPAPI on Windows, Keychain on macOS, libsecret on Linux) with encrypted persistence to a JSON file
- Harden `readSecureStorageFile` with JSON shape validation, null-prototype objects to prevent prototype pollution, and graceful fallback on corrupted data
- Use atomic temp-file + rename writes with `0o600` permissions to prevent data loss and interleaving
- Add GitHub OAuth browser web flow in the Electron main process (`github-oauth.ts`): opens the user's browser to GitHub's authorize URL, handles the callback via `headlamp://oauth/callback` deep link (or a temporary localhost server in dev mode), and exchanges the authorization code for tokens — all without routing through the backend `/externalproxy`
- Expose three IPC APIs via `desktopApi`: `startGitHubOAuth()`, `onGitHubOAuthCallback(cb)`, `refreshGitHubOAuth(refreshToken)` — token refresh runs in the main process to avoid CORS issues
- Register `headlamp://` as the default protocol client and handle deep links on both `second-instance` (warm start) and `open-url` (macOS) events

## Test plan
- [x] Verify secure storage save/load/delete round-trips correctly
- [x] Verify corrupted secure-storage.json (e.g. `null`, array, non-string values) returns an empty null-prototype object
- [x] Verify dangerous prototype keys (`__proto__`, `constructor`, `prototype`) are filtered out
- [x] Verify OAuth callback exchanges code for tokens and sends result to renderer
- [x] Verify CSRF state mismatch is rejected
- [x] Verify token refresh via IPC returns new tokens on success and error on failure
- [x] Verify dev callback server starts and handles a single request
- [x] All 68 Electron tests pass (`npm test` in `app/`)

Related PR:
- https://github.com/Azure/aks-desktop/pull/261

### DIFF
--- a/app/electron/github-oauth.ts
+++ b/app/electron/github-oauth.ts
@@ -1,0 +1,327 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Portions (c) Microsoft Corp.
+
+// aksd: GitHub OAuth web flow — handles auth start, deep-link callback, and token refresh
+// entirely in the Electron main process to avoid CORS issues and eliminate the backend proxy.
+//
+// In production the callback arrives via the `headlamp://oauth/callback` custom URL scheme.
+// In dev mode (ELECTRON_DEV) we spin up a temporary localhost HTTP server instead, because
+// custom URL schemes are not registered during development (especially under WSL2).
+
+import crypto from 'crypto';
+import type { BrowserWindow } from 'electron';
+import { app, ipcMain, shell } from 'electron';
+import http from 'http';
+import path from 'path';
+import { GITHUB_OAUTH_CALLBACK, GITHUB_OAUTH_REFRESH, GITHUB_OAUTH_START } from './ipc-channels';
+import { handleSecureStorageSave } from './secure-storage';
+
+export const CLIENT_ID = 'Iv23liWWbvrfIrA6WWj5';
+// The client secret is embedded intentionally. Native/desktop apps are "public clients" that
+// cannot keep secrets — any user can extract them from the binary. Per RFC 8252 §8.5
+// (https://datatracker.ietf.org/doc/html/rfc8252#section-8.5), secrets shipped in distributed
+// native apps SHOULD NOT be treated as confidential. Security relies on the registered redirect
+// URI, not the secret. GitHub Desktop, VS Code, and other Electron apps follow this pattern.
+export const CLIENT_SECRET = '5a066cc6ca8d2c6f201c45f24f2f4e62905b5d95';
+export const REDIRECT_URI = 'headlamp://oauth/callback';
+export const DEV_CALLBACK_PORT = 48321;
+export const DEV_REDIRECT_URI = `http://localhost:${DEV_CALLBACK_PORT}/oauth/callback`;
+export const STORAGE_KEY = 'aks-desktop:github-auth';
+export const TOKEN_URL = 'https://github.com/login/oauth/access_token';
+export const AUTH_URL = 'https://github.com/login/oauth/authorize';
+
+/** Shape of the JSON response from GitHub's OAuth token endpoint. */
+interface GitHubTokenResponse {
+  access_token?: string;
+  refresh_token?: string;
+  expires_in?: number;
+  error?: string;
+  error_description?: string;
+}
+
+/** Module-scoped pending CSRF state (single pending flow at a time). */
+let pendingState: string | null = null;
+
+/** Visible for testing — returns the current pending state. */
+export function getPendingState(): string | null {
+  return pendingState;
+}
+
+/**
+ * Visible for testing — sets the pending state.
+ * @param state - CSRF state string to store, or `null` to clear.
+ */
+export function setPendingState(state: string | null): void {
+  pendingState = state;
+}
+
+/**
+ * Saves token data to encrypted secure storage.
+ * @param accessToken - GitHub OAuth access token.
+ * @param refreshToken - GitHub OAuth refresh token.
+ * @param expiresAt - ISO-8601 timestamp when the access token expires.
+ */
+function saveTokens(accessToken: string, refreshToken: string, expiresAt: string): void {
+  const storagePath = path.join(app.getPath('userData'), 'secure-storage.json');
+  const result = handleSecureStorageSave(
+    storagePath,
+    STORAGE_KEY,
+    JSON.stringify({ accessToken, refreshToken, expiresAt })
+  );
+  if (!result.success) {
+    console.error('Failed to persist tokens to secure storage:', result.error);
+  }
+}
+
+/** Active dev callback server, if any. */
+let devServer: http.Server | null = null;
+
+/** Visible for testing — returns the active dev server. */
+export function getDevServer(): http.Server | null {
+  return devServer;
+}
+
+/** Stops the dev callback server if running. */
+function stopDevServer(): void {
+  if (devServer) {
+    devServer.close();
+    devServer = null;
+  }
+}
+
+/**
+ * Starts a temporary localhost HTTP server that handles a single OAuth callback,
+ * then shuts itself down. Used in dev mode where custom URL schemes aren't registered.
+ */
+function startDevCallbackServer(getMainWindow: () => BrowserWindow | null): void {
+  stopDevServer();
+
+  const server = http.createServer(async (req, res) => {
+    if (!req.url?.startsWith('/oauth/callback')) {
+      res.writeHead(404);
+      res.end('Not found');
+      return;
+    }
+
+    const urlObj = new URL(req.url, `http://127.0.0.1:${DEV_CALLBACK_PORT}`);
+    const mainWindow = getMainWindow();
+
+    if (!mainWindow) {
+      res.writeHead(503);
+      res.end('Main window not available');
+      return;
+    }
+
+    await handleOAuthCallback(urlObj, mainWindow, DEV_REDIRECT_URI);
+
+    res.writeHead(200, { 'Content-Type': 'text/html' });
+    res.end(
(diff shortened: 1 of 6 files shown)


app/electron/secure-storage.ts
```diff
@@ -0,0 +1,126 @@
...
+export const DANGEROUS_KEYS = new Set(['__proto__', 'constructor', 'prototype']);
+export const MAX_KEY_LENGTH = 128;
+export const MAX_VALUE_LENGTH = 64 * 1024; // 64 KB
+export const MAX_ENTRIES = 256;
+
+export function isValidKey(key: string): boolean {
+  return VALID_KEY_RE.test(key) && key.length <= MAX_KEY_LENGTH && !DANGEROUS_KEYS.has(key);
+}
+
+export function readSecureStorageFile(storagePath: string): Record<string, string> {
```
illume, line 41: These functions could do with tsdoc documentation.

app/electron/main.ts
```diff
@@ -1816,6 +1821,41 @@ async function startElecron() {
       }
     });
 
+    // aksd: Secure storage via Electron safeStorage API
```
illume, line 1859: So that there's less happening in the already big main.ts... we're trying to move things out of main.ts as we go.

All of this could could go into a separate module. Then main.ts could call that code.

app/electron/secure-storage.test.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Portions (c) Microsoft Corp.
+
+import { afterEach, beforeEach, describe, expect, it } from '@jest/globals';
```
illume, line 19: We're using vitest for tests.

Makefile
```diff
@@ -230,10 +230,10 @@ run-backend:
 	@echo "**** Warning: Running with Helm and dynamic-clusters endpoints enabled. ****"
 
 ifeq ($(UNIXSHELL),true)
-	HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/* -listen-addr=localhost
+	HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/*,https://github.com/login/* -listen-addr=localhost
```
illume: I'm not clear on what this is for.

In the "aksd: Add GitHub OAuth proxy-urls to app-build-manifest" commit message it could do with a Why, perhaps describing what this is for.

app/electron/secure-storage.ts
```diff
@@ -0,0 +1,126 @@
...
+    return { success: false, error: 'Value too large or invalid' };
+  }
+  if (!safeStorage.isEncryptionAvailable()) {
+    return { success: false, error: 'encryption-unavailable' };
+  }
+  const store = readSecureStorageFile(storagePath);
+  if (!(key in store) && Object.keys(store).length >= MAX_ENTRIES) {
+    return { success: false, error: 'Storage entry limit reached' };
+  }
+  const encrypted = safeStorage.encryptString(value).toString('base64');
```
illume: I don't know if you considered encryptStringAsync?

Makefile
illume: I wonder if we already know what specific things are needed under login/? Can we reduce it further?

app/electron/preload.ts
```diff
@@ -89,4 +89,22 @@ contextBridge.exposeInMainWorld('desktopApi', {
   },
 
   platform: process.platform,
+
+  // aksd: Secure storage — encrypt/decrypt via Electron safeStorage
+  secureStorageSave: (
```
illume, line 104: It might be worth importing these from another module and injecting them as well.

app/electron/secure-storage.test.ts
gambtho: it looks like the app/electron/ tests (plugin-management.test.ts, runCmd.test.ts) all use @jest/globals and the app's test script runs jest. Vitest is used in the plugin directory?

app/electron/secure-storage.ts
gambtho: is that on the electron safestorage api? i only saw the sync versions -- i think for this we shouldn't see any perf issues, but maybe future use?

Makefile
gambtho: The device flow hits github.com/login/device/code and github.com/login/oauth/access_token. We could narrow to those two exact paths, but login/* is already scoped to just the login subtree -- can tighten if you think its needed?

app/electron/preload.ts
gambtho: The channel names are shared via secure-storage-channels.ts. The preload functions themselves are one-liner ipcRenderer.invoke() wrappers — extracting them felt like it would add a file without much benefit. Happy to move them if you feel strongly about it though.

app/electron/secure-storage.test.ts
sniok: electron app still uses jest, we haven't migrated to vitest here yet

illume: ah yes, my bad.

app/electron/preload.ts
illume: No I don't feel strongly about it.

app/electron/secure-storage.ts
illume: This is mainly for storing GITHUB TOKEN? 

I thought generating keys and such generally burns CPU in order to make it secure... and if that is on the main thread it might block things for a while. If you didn't notice a pause there's probably no problem.

How about we see how it is when it's hooked up to the UI? I'll mark this as resolved.

Makefile
illume: Might be a good idea so we don't need to consider everything else under /login.

I saw at least "GET https://github.com/login/oauth/authorize" in there... https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps

I don't know what else there is really, so if you know how to tighten it a bit more, I think it could be worth doing.

Makefile
```diff
@@ -228,12 +228,13 @@ frontend-build-storybook:
 
 run-backend:
 	@echo "**** Warning: Running with Helm and dynamic-clusters endpoints enabled. ****"
-
+# github.com/login/* is proxied so the frontend can complete GitHub OAuth device-flow
```
illume: I don't really understand why it's needed? Can you please expand a little?

(I'm a little worried about the security implications of opening this up.)

illume (commented):
~~I'm missing some context (like an issue or related PR maybe?) on how we're going to use these two things?~~

~~Is there a future PR coming which uses these?~~

Edit: Ok, I see there is a related PR here:
- #261

illume:
Sorry, I missed the related PR... let me review again. 
- https://github.com/Azure/aks-desktop/pull/261

app/electron/secure-storage.ts
gambtho: These were added in a subsequent push — each exported function now has a JSDoc comment describing its purpose.

Makefile
gambtho: The GitHub OAuth device flow requires two HTTP requests from the frontend: `POST github.com/login/device/code` (to initiate the flow and get a user code) and `POST github.com/login/oauth/access_token` (to poll for the token after the user authorizes). Electron's renderer process can't make these cross-origin requests directly due to CORS, so they go through Headlamp's backend `/externalproxy` endpoint, which only allows URLs matching the `-proxy-urls` allowlist. Without this, the plugin would need to shell out to `curl` for these requests.

I've tightened the scope from `login/*` to just `login/device/*` and `login/oauth/*` to minimize the surface area.

Makefile
gambtho: done :)

app/electron/github-oauth.ts
```diff
@@ -0,0 +1,313 @@
...
+export function getPendingState(): string | null {
+  return pendingState;
+}
+
+/** Visible for testing — sets the pending state. */
+export function setPendingState(state: string | null): void {
+  pendingState = state;
+}
+
+/** Saves token data to encrypted secure storage. */
```
illume: Please document the arguments?

app/electron/github-oauth.ts
```diff
@@ -0,0 +1,313 @@
...
+/** Module-scoped pending CSRF state (single pending flow at a time). */
+let pendingState: string | null = null;
+
+/** Visible for testing — returns the current pending state. */
+export function getPendingState(): string | null {
+  return pendingState;
+}
+
+/** Visible for testing — sets the pending state. */
+export function setPendingState(state: string | null): void {
```
illume, line 69: Please document the argument?

app/electron/secure-storage.ts
```diff
@@ -0,0 +1,189 @@
...
+export const MAX_VALUE_LENGTH = 64 * 1024; // 64 KB
+export const MAX_ENTRIES = 256;
+
+/** Checks whether a storage key is safe to use: alphanumeric with colons/hyphens/underscores, within length limit, and not a dangerous prototype property. */
+export function isValidKey(key: string): boolean {
+  return VALID_KEY_RE.test(key) && key.length <= MAX_KEY_LENGTH && !DANGEROUS_KEYS.has(key);
+}
+
+/** Reads and validates the encrypted storage JSON file. Returns a null-prototype object of string entries, falling back to an empty null-prototype object on missing/corrupt files. */
+export function readSecureStorageFile(storagePath: string): Record<string, string> {
```
illume, line 41: Please document the function argument?

app/electron/secure-storage.ts
```diff
@@ -0,0 +1,189 @@
...
+        result[k] = v;
+      }
+    }
+    return result;
+  } catch {
+    return Object.create(null);
+  }
+}
+
+/** Atomically writes the storage object to disk using a temp-file + rename pattern with `0o600` permissions. */
```
illume: Please document the function argument?

app/electron/secure-storage.ts
```diff
@@ -0,0 +1,189 @@
...
+    try {
+      fs.unlinkSync(tmpPath);
+    } catch {
+      // Best-effort cleanup
+    }
+    throw error;
+  }
+}
+
+/** Encrypts `value` with Electron safeStorage and persists it under `key`. Enforces key/value validation, entry limits, and encryption availability. */
```
illume: Please document the function arguments?

PR #261, opened 2026-02-17
feat: add GitHub OAuth device flow auth, API wrapper, and secure storage

gambtho:
## Summary
- Add GitHub OAuth device flow authentication using Headlamp's `/externalproxy` to bypass CORS on GitHub's token endpoints
- Implement device code initiation, token exchange polling, and refresh token support
- Add Electron `safeStorage` wrapper for encrypted token persistence (no insecure fallback)
- Add Octokit-based GitHub API wrapper for repository listing, workflow dispatch, and file operations
- Define GitHub repo and workflow types

### Key design decisions
- **Proxy over curl**: Routes OAuth requests through Headlamp's backend `/externalproxy` endpoint instead of shelling out to `curl`, avoiding command allowlist changes, cross-platform issues, and process-list token exposure
- **Secure storage validation**: `getDesktopApi()` validates all three IPC methods before returning the API to prevent partial-implementation runtime errors
- **Token expiry safety**: Invalid/unparsable expiry timestamps are treated as expired to prevent indefinite use of bad tokens

## Test plan
- [x] `github-auth` unit tests pass (26 tests)
- [x] `github-api` unit tests pass (36 tests)
- [x] TypeScript type check passes
- [x] ESLint + Prettier pass
- [x] Requires `proxy-urls` in `app-build-manifest.json` (PR #260)

### DIFF
--- a/plugins/aks-desktop/src/utils/github/github-auth.ts
+++ b/plugins/aks-desktop/src/utils/github/github-auth.ts
@@ -5,8 +5,8 @@ import { secureStorageDelete, secureStorageLoad, secureStorageSave } from './sec
 
 // GitHub App public client ID and slug — safe to embed in desktop app
 // TODO: Replace with production app slug before release
-export const GITHUB_APP_CLIENT_ID = 'Iv23liWWbvrfIrA6WWj5';
-export const GITHUB_APP_SLUG = 'aks-desktop-testing';
+const GITHUB_APP_CLIENT_ID = 'Iv23liWWbvrfIrA6WWj5';
+const GITHUB_APP_SLUG = 'aks-desktop-testing';
 export const GITHUB_APP_INSTALL_URL = `https://github.com/apps/${GITHUB_APP_SLUG}/installations/new`;
 
 const DEVICE_CODE_URL = 'https://github.com/login/device/code';
@@ -51,7 +51,7 @@ function requireNumber(data: Record<string, unknown>, key: string): number {
  * so browser fetch() is blocked. Routing through the backend proxy avoids CORS.
  * Requires the target URL to be allowlisted in app-build-manifest.json proxy-urls.
  */
-export const githubOAuthPost = async (
+const githubOAuthPost = async (
   url: string,
   params: Record<string, string>
 ): Promise<Record<string, unknown>> => {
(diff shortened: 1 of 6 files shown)


plugins/aks-desktop/src/utils/github/github-auth.ts
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * Persists tokens using Electron safeStorage (OS-level encryption).
+ * If secure storage is unavailable, tokens are not persisted — they remain
+ * usable for the current session but will require re-authentication next time.
+ */
+export const saveTokens = async (tokens: StoredTokens): Promise<void> => {
+  const json = JSON.stringify(tokens);
+  const saved = await secureStorageSave(STORAGE_KEY, json);
+  if (!saved) {
```
sniok: should we fallback to an unsecure storage? I think if secure save fails then we shouldn't proceed

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -29,72 +27,21 @@ function needsRelogin(error: string): boolean {
...
-  return await runCommandAsync(command, args);
+function runKubectlCommand(args: string[]): Promise<{ stdout: string; stderr: string }> {
+  debugLog('[KUBECTL] Executing command: kubectl with args:', args);
+  return runCommandAsync('kubectl', args);
 }
 
 export async function isAzCliInstalled(): Promise<boolean> {
   try {
-    const { stdout, stderr } = await tryExec('az', ['version']);
+    const { stdout, stderr } = await runAzCommand(['version']);
```
sniok: renaming tryExec to runAzCommand has created quite a large diff since it needs to be updated in many spots. maybe we could just keep the old name and arguments?

plugins/aks-desktop/src/utils/github/github-auth.ts
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache 2.0.
+
+import { secureStorageDelete, secureStorageLoad, secureStorageSave } from './secure-storage';
+
+// GitHub App public client ID and slug — safe to embed in desktop app
+// TODO: Replace with production app slug before release
+export const GITHUB_APP_CLIENT_ID = 'Iv23liWWbvrfIrA6WWj5';
+export const GITHUB_APP_SLUG = 'aks-desktop-testing';
```
sniok: GITHUB_APP_CLIENT_ID and GITHUB_APP_SLUG are not used outside this file so they shouldn't be exported

plugins/aks-desktop/src/utils/github/github-auth.ts
```diff
@@ -0,0 +1,249 @@
...
+  return val;
+}
+
+/**
+ * Makes a POST request to a GitHub OAuth endpoint via Headlamp's /externalproxy.
+ * GitHub's OAuth endpoints don't return Access-Control-Allow-Origin headers,
+ * so browser fetch() is blocked. Routing through the backend proxy avoids CORS.
+ * Requires the target URL to be allowlisted in app-build-manifest.json proxy-urls.
+ */
+export const githubOAuthPost = async (
```
sniok: this function is not used outside this file, export is not necessary

plugins/aks-desktop/src/utils/azure/az-cli.ts
gambtho: that shouldn't be in this pr at all, removing -- was part of the earlier pr and removed

plugins/aks-desktop/src/utils/github/github-auth.ts
gambtho: i think this caused an issue in the dev env, will validate

PR #325, opened 2026-02-26
Update app and plugin versions to 0.2.0-alpha

yolossn:
## Description

Update aksd package version to 0.2.0-alpha

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aks-desktop-monorepo",
-  "version": "0.1.0-alpha",
+  "version": "0.2.0-alpha",
   "description": "AKS desktop - Headlamp Plugin Monorepo",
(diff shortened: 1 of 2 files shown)


plugins/aks-desktop/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "aks-desktop",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
illume: 0.2.0?

0.1.1 would be a patch release

package.json
illume: 0.2.0-alpha ?

0.1.1 would be a patch release

skoeva: agreed 0.2.0-alpha would follow convention

illume (approved):

PR #329, opened 2026-02-26
app: Github OAuth: migrate tests to vitest from jest

sniok:
update jest API to their vitest equivalents
replace require() calls with static ESM imports to fix module resolution failures under vitest


illume (approved):
Thanks!

PR #332, opened 2026-02-26
app: Revert electron-builder version bump

sniok:
this pr reverts electron-builder version bump and build configuration


illume (approved):
🎉 thanks!

PR #368, opened 2026-03-05
headlamp: Update submodule to the latest commit

sniok:
just bumping submodule to the latest headlamp-downstream commit


skoeva (approved):
lgtm

PR #412, opened 2026-03-11
ci: Check if translations are up to date

sniok:
Add a CI check for translation in plugins/aks-desktop/


sniok:
check works (current translations are outdated) https://github.com/Azure/aks-desktop/actions/runs/22964452966/job/66663989536?pr=412

we can merge this after https://github.com/Azure/aks-desktop/pull/411

PR #457, opened 2026-03-17
aksd: derive deep link protocol scheme from package.json

### LINKED ISSUE #458: GitHub OAuth callback opens Headlamp instead of AKS Desktop
## Problem

When a user completes the "Connect GitHub" OAuth flow and has both Headlamp and AKS Desktop installed, the OAuth callback opens Headlamp instead of AKS Desktop.

## Root Cause

The deep link protocol scheme was hardcoded as `headlamp://` in the Electron main process. Both AKS Desktop and Headlamp registered the same `headlamp://` protocol, so the OS could route the `headlamp://oauth/callback` redirect to the wrong app.

## Fix

- The protocol scheme is now read dynamically from `package.json`'s `build.protocols.schemes` at startup, so each Headlamp-based app uses its own registered scheme.
- AKS Desktop's scheme is set to `aks-desktop://` in its `package.json`.
- The GitHub OAuth App's callback URL must be updated to `aks-desktop://oauth/callback`.

## Required Configuration Change

The GitHub OAuth App (client ID `Iv23liWWbvrfIrA6WWj5`) needs its **Authorization callback URL** updated to `aks-desktop://oauth/callback` in the GitHub App settings.

gambtho:
## Summary

- Deep link protocol was hardcoded as `headlamp://`, causing OAuth callbacks to open Headlamp instead of AKS Desktop when both are installed
- Protocol scheme is now read dynamically from `build.protocols.schemes` in `package.json` at startup, so each Headlamp-based app automatically uses its own registered scheme
- AKS Desktop's scheme updated from `headlamp` to `aks-desktop`

Fixes #458

## Changes

- **`app/package.json`** — Protocol scheme changed from `headlamp` to `aks-desktop`
- **`app/electron/main.ts`** — Added `getProtocolScheme()` that reads the scheme from `package.json` at runtime; all deep link handling uses the dynamic value
- **`app/electron/github-oauth.ts`** — Removed hardcoded `REDIRECT_URI`; `setupGitHubOAuthHandlers` accepts `protocolScheme` option and builds the redirect URI dynamically
- **`app/electron/github-oauth.test.ts`** — Tests updated to use a generic `test-app://` scheme and pass redirect URI explicitly

## Github App Changes

The GitHub OAuth App (client ID `Iv23liWWbvrfIrA6WWj5`) had its callback URL updated to include `aks-desktop://oauth/callback` in the GitHub App settings, and was transferred to the Azure Org

## Testing

- [x] `npm run tsc` — type check passes
- [x] `npm test` — all 73 tests pass
- [x] No remaining hardcoded `headlamp://` references in `app/electron/`
- [ ] End-to-end: build app, trigger "Connect GitHub", confirm callback opens AKS Desktop

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -57,6 +57,26 @@ import { addRunCmdConsent, removeRunCmdConsent, runScript, setupRunCmdHandlers }
 import { setupSecureStorageHandlers } from './secure-storage';
 import windowSize from './windowSize';
 
+/**
+ * Reads the custom URL protocol scheme from package.json's build.protocols.schemes.
+ * Falls back to the app name (lowercased, spaces replaced with hyphens), then 'headlamp'.
+ */
+function getProtocolScheme(): string {
+  try {
+    const pkgPath = path.join(__dirname, '..', 'package.json');
+    const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+    const schemes: string[] | undefined = pkg?.build?.protocols?.schemes;
+    if (schemes && schemes.length > 0) {
+      return schemes[0];
+    }
+  } catch {
+    // Fall through to fallback
+  }
+  return app.name?.toLowerCase().replace(/\s+/g, '-') || 'headlamp';
+}
+
+const protocolScheme = getProtocolScheme();
+
 let isRunningScript = false;
 if (process.env.HEADLAMP_RUN_SCRIPT) {
   isRunningScript = true;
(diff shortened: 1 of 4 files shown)


app/electron/main.ts
```diff
@@ -57,6 +57,26 @@ import { addRunCmdConsent, removeRunCmdConsent, runScript, setupRunCmdHandlers }
 import { setupSecureStorageHandlers } from './secure-storage';
 import windowSize from './windowSize';
 
+/**
+ * Reads the custom URL protocol scheme from package.json's build.protocols.schemes.
+ * Falls back to the app name (lowercased, spaces replaced with hyphens), then 'headlamp'.
+ */
+function getProtocolScheme(): string {
```
illume, line 64: @gambtho did you test this in the compiled app? I'm not sure if the path is correct.

gambtho: i'm blocked in testing it until #407 merges, as the transfer to azure included changing the redirect url

gambtho: electron-builder always includes package.json at the app root, and the __dirname of the bundled build/main.js is one level below it -- so, it should work? @illume

illume: Ok, cool.

PR #468, opened 2026-03-18
Deploy: Add secrets, workload identity, ACR support, and fix identity role assignments

### LINKED ISSUE #467: Add secrets support and workload identity to Deploy application flow
## Summary

Add two features to the Deploy Wizard's Container configuration path:

### 1. Secrets Support
Allow users to mark individual environment variables as "secret" so that:
- Values are masked in the UI (password-style input)
- They are stored as a Kubernetes `Secret` resource (not inline in the Deployment manifest)
- Referenced via `valueFrom.secretKeyRef` in the Deployment's env section

Currently all env vars are embedded as plain text in the generated Deployment manifest, which is undesirable for sensitive values like API keys.

### 2. Workload Identity Integration
Add a "Workload Identity" step to the container configuration stepper that:
- Provides a toggle to enable Azure Workload Identity
- Lets the user **create a new** or **select an existing** user-assigned managed identity
- Creates federated credentials for the AKS cluster's OIDC issuer + namespace + service account
- Generates a `ServiceAccount` resource with `azure.workload.identity/client-id` annotation
- Adds `azure.workload.identity/use: "true"` label to the pod template
- Sets `serviceAccountName` on the pod spec

### 3. GitHub Pipeline Integration
Both features need to be reflected in the generated GitHub Pipeline:
- Agent template (`agentTemplates.ts`) should include workload identity instructions (ServiceAccount manifest, pod labels)
- Workflow instructions should create the ServiceAccount before applying manifests
- Env var secret handling remains compatible (all env vars already become GH secrets)

## Key Files
- `plugins/aks-desktop/src/components/DeployWizard/hooks/useContainerConfiguration.ts` — ContainerConfig type
- `plugins/aks-desktop/src/components/DeployWizard/components/ConfigureContainer.tsx` — UI form
- `plugins/aks-desktop/src/components/DeployWizard/utils/yamlGenerator.ts` — YAML generation
- `plugins/aks-desktop/src/components/DeployTab/utils/extractContainerConfig.ts` — Edit round-tripping
- `plugins/aks-desktop/src/utils/azure/az-cli.ts` — Azure CLI integration
- `plugins/aks-desktop/src/components/GitHubPipeline/utils/agentTemplates.ts` — Pipeline agent config

## Motivation
Users deploying apps that integrate with Azure services (e.g., Azure OpenAI) need to pass API keys without exposing them as plain text. Workload Identity provides a more secure, keyless alternative for Azure service authentication.

gambtho:
## Summary

- Add per-variable secret toggle in the Deploy Wizard environment variables step — secret values are masked in the UI, stored as a Kubernetes `Secret` resource, and referenced via `valueFrom.secretKeyRef` in the Deployment manifest
- Add a new "Workload Identity" stepper step that creates (or selects existing) Azure managed identities, configures federated credentials against the cluster's OIDC issuer, and generates a `ServiceAccount` with `azure.workload.identity/client-id` annotation plus pod labels/serviceAccountName on the Deployment
- Update GitHub Pipeline agent templates to include workload identity ServiceAccount creation and pod configuration in generated workflows
- Fix managed identity role assignments: assign the full required role matrix (AcrPush, Container Registry Tasks Contributor, AKS RBAC Writer, AKS Namespace User) scoped correctly for both normal and managed namespaces
- Add ACR selection/creation to the pipeline wizard so the generated workflow can build and push container images
- Extract shared identity setup logic into reusable utilities, eliminating duplication between the GitHub Pipeline and Deploy Wizard flows

## Type of Change

## Related Issues

Closes #467

### Secrets support
- Extended `ContainerConfig.envVars` type with `isSecret: boolean`
- Added lock icon toggle and password masking in `ConfigureContainer.tsx`
- Updated `yamlGenerator.ts` to emit a `Secret` resource and `secretKeyRef` entries for secret env vars
- Updated `extractContainerConfig.ts` to round-trip `secretKeyRef` entries

### Workload Identity
- Added `getAksOidcIssuerUrl`, `listManagedIdentities`, `createK8sFederatedCredential` to `az-cli.ts`
- Created `useDeployWorkloadIdentity` hook (adapts existing GitHub Pipeline pattern)
- Added new stepper step with "Create new" / "Use existing" identity modes
- Updated `yamlGenerator.ts` to emit `ServiceAccount`, pod label, and `serviceAccountName`
- Threaded Azure context from `ClusterDeployCard` through `DeployWizard` to `ConfigureContainer`

### Identity role assignment fixes
- Added `assignRolesToIdentity()` — generalized multi-role assignment function replacing the single hardcoded role
- Added `computeRequiredRoles()` in new `identityRoles.ts` — computes the correct role set based on namespace type (normal vs managed), ACR presence, and Azure RBAC status
- **Normal namespace roles:** AcrPush (ACR), Container Registry Tasks Contributor (ACR), AKS Cluster User Role (cluster), AKS RBAC Writer (cluster, if Azure RBAC enabled)
- **Managed namespace roles:** AcrPush (ACR), Container Registry Tasks Contributor (ACR), AKS RBAC Writer (MNS scope), AKS Namespace User (MNS scope)
- Extracted `ensureIdentityAndResourceGroup()` into shared `identitySetup.ts` — eliminates ~80% duplication between the two identity hooks
- Added `getManagedNamespaceResourceId()` and `buildClusterScope()` helpers to `az-cli.ts`

### ACR support in pipeline wizard
- Added `AcrSelector` component — lists existing ACRs, supports creating new registries, shows a warning if skipped
- Added `createContainerRegistry()` to `az-cli.ts`
- Added `acrResourceId` and `acrLoginServer` to `PipelineConfig`
- Agent templates now include `az acr build` instructions when ACR is configured
- Pipeline secrets now include `AZURE_ACR_NAME` when ACR is selected

### Pipeline integration
- Added `generateWorkloadIdentityWorkflowInstructions` to `agentTemplates.ts`
- Included workload identity config summary in generated agent markdown

## Testing

- [x] Unit tests pass
- [x] Manual testing completed

### Test Cases

1. All 208 related tests pass (identity roles, identity setup, pipeline hooks, pipeline components, agent templates, pipeline orchestration)
2. TypeScript type check passes
3. ESLint and Prettier pass
4. No new test failures introduced
5. New test files: `identityRoles.test.ts` (10 tests), `identitySetup.test.ts` (8 tests)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

## Performance Impact

- [x] No performance impact

## Reviewer Notes

- The workload identity step is disabled when Azure context is unavailable (non-Azure clusters or not signed in)
- Secret values are visible in the YAML preview step (acceptable since this is a local desktop app and the user just typed them)
- The `Secret` resource is placed first in the generated YAML so it exists before the Deployment that references it
- ACR selection is optional with a warning — pipelines can still be created without an ACR for pre-built image scenarios
- The old `assignRoleToIdentity()` function is preserved for backward compatibility but should be removed once all callers are migrated
- `getClusterCapabilities()` already returns `azureRbacEnabled` — callers just need to thread it through to the identity hooks


plugins/aks-desktop/src/components/DeployWizard/utils/yamlGenerator.ts
```diff
@@ -15,11 +16,21 @@ export type ContainerDeploymentConfig = Omit<
...
- * Type with only the fields needed for YAML generation from ContainerConfigurationState
+ * Type alias for ContainerDeploymentConfig, used as the input for YAML generation.
  */
-export type ContainerConfigForYaml = Omit<ContainerDeploymentConfig, 'namespace'> & {
-  namespace?: string;
-};
+export type ContainerConfigForYaml = ContainerDeploymentConfig;
+
+/** Normalize a string to DNS-1123 label constraints (lowercase alnum/`-`, max 63 chars). */
+function normalizeK8sName(raw: string): string {
```
sniok: there's already a very similar function `sanitizeLabelValue` in `useDeploymentHealth.ts`

plugins/aks-desktop/src/components/DeployTab/utils/extractContainerConfig.ts
```diff
@@ -6,7 +6,11 @@ import type { ContainerConfig } from '../../DeployWizard/hooks/useContainerConfi
 interface KubeContainer {
   image?: string;
```
sniok: we have a similar type in headlamp https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/cluster.ts#L181

could we reuse it here?

plugins/aks-desktop/src/components/DeployWizard/components/ConfigureContainer.tsx
```diff
@@ -1013,10 +1080,18 @@ export default function ConfigureContainer({
           <StepContent>
             <Box sx={{ display: 'flex', flexDirection: 'column', gap: 1 }}>
               {containerConfig.config.envVars.map((pair, idx) => (
-                <Box key={idx} sx={{ display: 'flex', gap: 1 }}>
+                <Box key={idx} sx={{ display: 'flex', gap: 1, alignItems: 'center' }}>
                   <TextField
                     label={t('Key')}
                     value={pair.key}
+                    error={!!pair.key.trim() && !/^[a-zA-Z_][a-zA-Z0-9_]*$/.test(pair.key)}
```
sniok: this regex probably should be extracted and explained a bit

plugins/aks-desktop/src/components/DeployWizard/components/ConfigureContainer.tsx
sniok, line 110: this file is getting quite large, consider splitting it into multiple if possible

plugins/aks-desktop/src/components/DeployWizard/components/ConfigureContainer.tsx
```diff
@@ -1227,7 +1324,290 @@ export default function ConfigureContainer({
...
+                  }
+                }}
+              >
+                {t('Continue')}
+              </Button>
+            </Box>
+          </StepContent>
+        </Step>
+
+        {/* 8. Advanced */}
```
sniok: these kind of comments are probably not necessary

plugins/aks-desktop/src/components/DeployWizard/components/ConfigureContainer.tsx
```diff
@@ -1227,7 +1324,290 @@ export default function ConfigureContainer({
...
+                        containerConfig.config.workloadIdentityServiceAccount}
+                    </Typography>
+                  </Box>
+                )}
+              </Box>
+            )}
+            <Box sx={{ mt: 2, display: 'flex', gap: 1 }}>
+              <Button
+                variant="outlined"
+                onClick={() => containerConfig.setConfig(c => ({ ...c, containerStep: 5 }))}
```
sniok: containerStep is kind of a magic number right now, it's probably better to give them names/ids

plugins/aks-desktop/src/components/DeployWizard/hooks/useContainerConfiguration.ts
```diff
@@ -47,6 +47,9 @@ export interface ContainerConfig {
   runAsNonRoot: boolean;
   readOnlyRootFilesystem: boolean;
   allowPrivilegeEscalation: boolean;
+  enableWorkloadIdentity: boolean;
```
sniok, line 75: that is so many properties, maybe they could be grouped somehow
it'd be nice to have some documentation for this interface although most of the fields are self explanatory

plugins/aks-desktop/src/components/DeployWizard/hooks/useDeployWorkloadIdentity.ts
```diff
@@ -0,0 +1,136 @@
...
+  | 'creating-credential'
+  | 'done'
+  | 'error';
+
+export interface DeployWorkloadIdentityResult {
+  clientId: string;
+  serviceAccountName: string;
+}
+
+export interface UseDeployWorkloadIdentityReturn {
```
sniok: these types shouldn't be exported, they're not used outside this file

plugins/aks-desktop/src/components/DeployWizard/utils/yamlGenerator.ts
```diff
@@ -15,11 +16,21 @@ export type ContainerDeploymentConfig = Omit<
 };
 
 /**
- * Type with only the fields needed for YAML generation from ContainerConfigurationState
+ * Type alias for ContainerDeploymentConfig, used as the input for YAML generation.
  */
-export type ContainerConfigForYaml = Omit<ContainerDeploymentConfig, 'namespace'> & {
-  namespace?: string;
-};
+export type ContainerConfigForYaml = ContainerDeploymentConfig;
```
sniok: why is this type alias needed?  `ContainerDeploymentConfig` could be used directly

plugins/aks-desktop/src/components/DeployWizard/utils/yamlGenerator.ts
```diff
@@ -37,21 +48,39 @@ function escapeYamlDoubleQuoted(input: string): string {
...
+    ...plainEnvVars.map(
       e =>
         `            - name: ${e.key}
               value: "${escapeYamlDoubleQuoted(e.value)}"`
-    )
-    .join('\n');
+    ),
+    ...secretEnvVars.map(
+      e =>
+        `            - name: ${e.key}
```
sniok: generating yaml from template literals looks like a bad idea. it'd be tricky to debug and maintain this

we're already using `yaml` package which can format json to yaml.

plugins/aks-desktop/src/components/GitHubPipeline/components/AcrSelector.tsx
```diff
@@ -0,0 +1,274 @@
...
+    setSelectedValue(value ? value.acrResourceId : '');
+  }, [value]);
+
+  // Create form state
+  const [newRegistryName, setNewRegistryName] = useState('');
+  const [newRegistrySku, setNewRegistrySku] = useState<'Basic' | 'Standard' | 'Premium'>('Basic');
+  const [creating, setCreating] = useState(false);
+  const [createError, setCreateError] = useState<string | null>(null);
+
+  const fetchRegistries = useCallback(async () => {
```
sniok: useCallback + useEffect is unecessary, single useEffect is enough here

```
useEffect(() => {
   // fetch registries
}, [subscriptionId])
```

also we can't use `t` in dependencies because it is always recreated on each render resulting in a loop

plugins/aks-desktop/src/hooks/useNamespaceCapabilities.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Fetches namespace capabilities (managed namespace status, Azure RBAC).
+ * Return values: `undefined` = still loading or indeterminate, `true`/`false` = resolved.
+ * For `isManagedNamespace`: only set to `true` on definitive success; `undefined` if the API
+ * returns a non-success result (could be transient/auth issue); `false` only on thrown errors.
+ * For `azureRbacEnabled`: `undefined` if the cluster capabilities are unavailable or not set.
+ */
+export function useNamespaceCapabilities(
+  subscriptionId: string | undefined,
```
sniok: I'd recommend using named arguments here

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -1255,24 +1268,31 @@ export async function getContainerRegistries(subscriptionId: string): Promise<an
...
-    const registries = JSON.parse(stdout || '[]');
-    return registries.map((registry: any) => ({
-      id: registry.id,
-      name: registry.name,
-      resourceGroup: registry.resourceGroup,
-      loginServer: registry.loginServer,
-      location: registry.location,
-      sku: registry.sku?.name || 'Basic',
-    }));
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
sniok: we don't have this rule

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -3157,6 +3177,209 @@ export async function assignRoleToIdentity(options: {
...
+  success: boolean;
+  results: Array<{ role: string; scope: string; success: boolean; error?: string }>;
+}
+
+/**
+ * Assigns multiple Azure RBAC roles to a managed identity.
+ * Treats `RoleAssignmentExists` as success (idempotent).
+ * Roles are assigned sequentially to avoid Azure ARM rate-limiting (429s).
+ */
+export async function assignRolesToIdentity(options: {
```
sniok: this file is already quite large, maybe we can put new functions somewhere else, maybe grouped per feature cluster/namespace/roles/etc.

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -3157,6 +3177,209 @@ export async function assignRoleToIdentity(options: {
...
+  subscriptionId: string;
+  location: string;
+  sku?: 'Basic' | 'Standard' | 'Premium';
+}): Promise<{ success: boolean; id?: string; loginServer?: string; error?: string }> {
+  const { registryName, resourceGroup, subscriptionId, location, sku = 'Basic' } = options;
+
+  if (!isValidGuid(subscriptionId)) {
+    return { success: false, error: 'Invalid subscription ID format' };
+  }
+  if (!/^[a-zA-Z0-9]{5,50}$/.test(registryName)) {
```
sniok: in AcrSelector.tsx the regex is slightly different, with slightly different message

```js
const acrNamePattern = /^[a-z0-9]{5,50}$/;
if (!acrNamePattern.test(trimmed)) {
  setCreateError(t('Registry name must be 5-50 characters, lowercase alphanumeric only.'));
  return;
}
```

and in the input the helper text is also slightly different 

```tsx
helperText={t(
  'Globally unique name. Only alphanumeric characters, 5-50 characters.'
)}
```

and here error message is also similar but a bit different

```
must be 5–50 alphanumeric characters only
```

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -3157,6 +3177,209 @@ export async function assignRoleToIdentity(options: {
...
+    return {
+      success: false,
+      error: 'Invalid registry name: must be 5–50 alphanumeric characters only',
```
sniok: this might be out of scope for this PR but I think all the error messages in this file need to be translated

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -3283,3 +3506,219 @@ export async function searchAzureADUsers(
...
+
+  if (!isValidGuid(subscriptionId)) {
+    return { success: false, error: 'Invalid subscription ID format' };
+  }
+  if (!isValidAzResourceName(clusterName) || !isValidAzResourceName(resourceGroup)) {
+    return { success: false, error: 'Invalid cluster name or resource group format' };
+  }
+
+  const query =
+    '{issuerUrl: oidcIssuerProfile.issuerUrl, workloadIdentityEnabled: securityProfile.workloadIdentity.enabled}';
```
sniok: the query might need to be escaped differently for different platforms using `quoteForPlatform` although it needs to be tested first. Have you tested this on windows/linux?

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -3283,3 +3506,219 @@ export async function searchAzureADUsers(
...
+    const parsed = new URL(issuerUrl);
+    if (parsed.protocol !== 'https:') {
+      return { success: false, error: 'OIDC issuer URL must use HTTPS' };
+    }
+  } catch {
+    return { success: false, error: 'Invalid OIDC issuer URL format' };
+  }
+
+  // Validate namespace and serviceAccountName (DNS label: lowercase alphanumeric and hyphens, max 63 chars)
+  const K8S_LABEL_PATTERN = /^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$/;
```
sniok: this regex is already defined in other places, please try to reuse them

plugins/aks-desktop/src/utils/azure/az-cli.ts
```diff
@@ -3283,3 +3506,219 @@ export async function searchAzureADUsers(
...
+  if (!K8S_LABEL_PATTERN.test(serviceAccountName)) {
+    return { success: false, error: 'Invalid Kubernetes service account name format' };
+  }
+
+  const subject = `system:serviceaccount:${namespace}:${serviceAccountName}`;
+  const rawCredentialName = `K8sSA-${namespace}-${serviceAccountName}`;
+  let credentialName = rawCredentialName;
+  if (credentialName.length > 128) {
+    let h = 0;
+    for (let i = 0; i < rawCredentialName.length; i++) {
```
sniok: is this hash calculation? this could probably be extracted into a separate function

plugins/aks-desktop/src/utils/kubernetes/serviceAccountNames.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/**
+ * Derives a Kubernetes ServiceAccount name from an application name.
+ * Normalizes to lowercase alphanumeric + hyphens, max 63 chars.
+ */
+export function getServiceAccountName(appName: string): string {
```
sniok, line 10: this is almost identical to the function `normalizeK8sName` introduced in this PR

plugins/aks-desktop/src/components/DeployTab/utils/extractContainerConfig.ts
gambtho: fields don't quite line-up, would have to do some unsafe casing to use the headlamp type

plugins/aks-desktop/src/utils/azure/az-cli.ts
gambtho: will test, but don't think it's needed as the args  are passed as array, consistent with existing code using --query, tested on linux, will try on windows

plugins/aks-desktop/src/components/DeployTab/utils/extractContainerConfig.ts
gambtho: actually, maybe this approach makes more sense?

plugins/aks-desktop/src/utils/azure/az-cli.ts
gambtho: validated

PR #503, opened 2026-03-21
InfoTab: Add caching for namespace details

### LINKED ISSUE #280: Info Tab: Doesn't load, infinite loop
If you go to project -> info tab it doesn't load, and there are many network requests 

related: #119

skoeva:
The InfoTab was making two sequential Azure CLI calls on every open, causing ~1 minute load times. This change replaces that with a single fetch backed by a module-level cache so subsequent opens are instant.

Fixes: #280 

### Summary

- Added a module-level `detailsCache` (stale-while-revalidate): cached data is shown immediately, a background fetch keeps it fresh. Cache key includes `subscription`/`resourceGroup`/`clusterName`/`projectId` to prevent cross-subscription collisions
- Revalidation is based on whether data is already displayed (not cache presence), so background fetches are used even after a save invalidates the cache — no cold-start spinner after saving
- Cache is invalidated after a successful save so stale data is never shown post-update
- `handleRefresh` triggers a background revalidation without clearing the cache, keeping the form visible
- Added `revalidating` state for background fetch indication
- Updated `InfoTab.tsx` to show a background spinner and a Refresh button
- Updated tests: removed obsolete `getManagedNamespaces` mock/cases; added coverage for cache hit, background revalidation, error suppression with existing data, cache invalidation on save, and post-save refresh running as revalidation; replaced direct `detailsCache` access with exported test helpers

### Testing
- [x] Run `cd plugins/aks-desktop && npm test` and ensure the tests pass


tejhan (dismissed):
Changes LGTM, tests pass & was able to replicated new behavior.

illume (changes requested):
@skoeva it seems there's a git conflict now. Can you please take a look?

illume (changes requested):
There's some git conflicts now. Can you please have a look?

(Also some review comment(s) left open)

skoeva:
we can use react query here instead of bespoke logic

PR #521, opened 2026-03-25
Enforce LF line endings for shell scripts

### LINKED ISSUE #520: [BUG] running `npm run plugin:setup` on windows fails
here's the output:

```
> insights-plugin@0.2.4 build
> bash scripts/download-artifacts.sh

: invalid option namefacts.sh: line 4: set: pipefail
node:internal/errors:932
  const err = new Error(message);
              ^

Error: Command failed: npm install && npm run build
    at __node_internal_genericNodeError (node:internal/errors:932:15)
    at checkExecSyncError (node:child_process:890:11)
    at execSync (node:child_process:962:15)
    at <anonymous> (c:\Users\eskopelitis\Development\aksd-sandbox\build\setup-plugins.ts:68:3)
    at Object.<anonymous> (c:\Users\eskopelitis\Development\aksd-sandbox\build\setup-plugins.ts:92:72)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Object.transformer (C:\Users\eskopelitis\Development\aksd-sandbox\node_modules\tsx\dist\register-D46fvsV_.cjs:3:1104)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at cjsLoader (node:internal/modules/esm/translators:356:17) {
  status: 2,
  signal: null,
  output: [ null, null, null ],
  pid: 22508,
  stdout: null,
  stderr: null
}

Node.js v20.11.1
```

the issue is related to line endings which can be fixed in the .gitattribute file

skoeva:
This change adds a `.gitattributes` file to ensure `.sh` files are always checked out with LF line endings, preventing `core.autocrlf=true` on Windows from converting them to CRLF. This fixes a build failure where bash would reject `set -o pipefail\r` as an invalid option.

Fixes: #520 

### Testing
- [x] Run `npm run plugin:setup` on Windows and ensure it doesn't fail

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf


tejhan (approved):
Tested & working now, LGTM.

illume (approved):
🎉 thanks!

PR #561, opened 2026-04-02
feat: RBAC fixes for pipeline identity — AcrPull on kubelet + AKS RBAC Writer

### LINKED ISSUE #550: Fast Path Pipeline: RBAC fixes (foundation)
Part of #557

## Problem

Two RBAC gaps cause deploy failures after the pipeline runs:

1. **Namespace annotation fails (F3):** When the deploy workflow tries to annotate the namespace and deployments, it fails with `User does not have access to the resource in Azure` because the pipeline's workload identity doesn't have `AKS RBAC Writer`. Today we only assign this role when `azureRbacEnabled` is detected on the cluster, but pipeline identities always need write access for annotations.

2. **Image pull fails (F1):** After the workflow successfully builds and pushes an image to ACR, pods fail with `ImagePullBackOff` because the AKS node pool's kubelet identity doesn't have `AcrPull` on the ACR. We assign `AcrPush` to the *pipeline* identity for building, but never grant pull access to the *kubelet* identity that actually runs the pods.

## Changes

- Add `isPipeline` flag to `IdentityRoleContext` — when true, always include `AKS RBAC Writer` regardless of cluster RBAC detection
- Wire `isPipeline: true` through `ensureIdentityWithRoles` for the GitHub pipeline flow
- Add `getKubeletIdentityObjectId()` — resolves the kubelet identity from `az aks show`
- Assign `AcrPull` to the kubelet identity on the ACR scope during pipeline identity setup
- Keep `continue-on-error: true` on workflow annotation steps as a safety net

gambtho:
## Summary

Foundation RBAC fixes for the fast path pipeline (and the existing agent path):

- **AcrPull on kubelet identity**: During pipeline setup, resolve the cluster's kubelet identity and assign `AcrPull` on the ACR scope so nodes can pull built images (fixes F1)
- **AKS RBAC Writer for pipeline identities**: Always assign `AKS RBAC Writer` for pipeline identities regardless of `azureRbacEnabled`, so the pipeline can write namespace/deployment annotations (fixes F3)
- Add `isPipeline` flag to `IdentityRoleContext` to distinguish pipeline identity setup from deploy wizard identity setup

## Changes Made

- `identityRoles.ts` — `computeRequiredRoles` now always includes `AKS RBAC Writer` when `isPipeline` is true
- `identityWithRoles.ts` — threading `isPipeline` flag through `ensureIdentityWithRoles`; kubelet AcrPull assignment returns warnings via result object (not status channel)
- `az-identity.ts` — new `getKubeletIdentityObjectId()` utility for resolving the kubelet identity's objectId
- `useWorkloadIdentitySetup.ts` — passes `isPipeline: true` to `ensureIdentityWithRoles`, which handles kubelet AcrPull assignment internally
- `useDeployWorkloadIdentity.ts` — passes `isPipeline: false` to preserve existing behavior
- Tests for all new logic

## Related Issues

Part of #557 (Fast Path Pipeline)
Closes #550

## Test Plan

- [x] Unit tests for `computeRequiredRoles` with `isPipeline` flag
- [x] Unit tests for `getKubeletIdentityObjectId`
- [x] Unit tests for `ensureIdentityWithRoles` with pipeline context (warnings via result object)
- [x] Type check passes
- [x] Existing tests unaffected


plugins/aks-desktop/src/utils/azure/az-identity.ts
```diff
@@ -245,6 +245,59 @@ export async function getManagedNamespaceResourceId(options: {
   return { success: true, resourceId: result.data };
 }
 
+/**
+ * Gets the kubelet identity's objectId for an AKS cluster.
+ * Used to assign AcrPull so nodes can pull container images.
```
illume: I'm not sure what AcrPull is? Could you please add a link to what this is here?

Also, I'm not sure what "kubelet identity's objectId for an AKS cluster" is either. If there's docs for that can you also please add this here?

plugins/aks-desktop/src/utils/azure/identityWithRoles.ts
```diff
@@ -119,5 +134,33 @@ export async function ensureIdentityWithRoles(
...
+      const kubeletRoleResult = await assignRolesToIdentity({
+        principalId: kubeletResult.objectId,
+        subscriptionId,
+        roles: [{ role: 'AcrPull', scope: acrResourceId }],
+      });
+      if (!kubeletRoleResult.success) {
+        warnings.push(
+          'Failed to assign AcrPull to kubelet identity. Nodes may not be able to pull images from ACR.'
+        );
+      }
```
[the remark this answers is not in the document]
illume: This seems valid.

plugins/aks-desktop/src/utils/azure/identityRoles.ts
```diff
@@ -61,7 +63,7 @@ export function computeRequiredRoles(ctx: IdentityRoleContext): RoleAssignment[]
     roles.push({ role: AKS_NAMESPACE_USER, scope: ctx.managedNamespaceResourceId });
   } else {
     roles.push({ role: AKS_CLUSTER_USER, scope: clusterScope });
-    if (ctx.azureRbacEnabled) {
+    if (ctx.azureRbacEnabled || ctx.isPipeline) {
```
illume: Just confirming this... I don't really understand the implications...

If azureRbacEnabled is false then it adds AKS_RBAC_WRITER?

What is this needed for exactly? I'm wondering if this is the minimum needed to achieve what's needed to be done here.

plugins/aks-desktop/src/components/GitHubPipeline/hooks/useWorkloadIdentitySetup.ts
```diff
@@ -87,10 +87,15 @@ export const useWorkloadIdentitySetup = (): UseWorkloadIdentitySetupReturn => {
         isManagedNamespace,
         namespaceName,
         azureRbacEnabled,
+        isPipeline: true,
         purpose: 'GitHub Actions Identity',
         onStatusChange: setStatus,
       });
 
+      if (identityResult.warnings.length > 0) {
+        console.warn('[WorkloadIdentitySetup] Warnings:', identityResult.warnings);
```
illume: Should this instead be presented to the user?

TODO: check if we collecting warnings in telemetry now, and what info could potentially leak here.

plugins/aks-desktop/src/utils/azure/az-identity.ts
```diff
@@ -245,6 +245,59 @@ export async function getManagedNamespaceResourceId(options: {
...
+    `Failed to get kubelet identity for cluster ${clusterName}`,
+    (stdout: string) => JSON.parse(stdout)
+  );
+
+  if (!result.success) {
+    return { success: false, error: result.error ?? 'Failed to get cluster details' };
+  }
+
+  const objectId = result.data;
+  if (!objectId || typeof objectId !== 'string') {
```
illume, line 317: Since there's already a validation function. I think validating the data here could be a good idea (just to be cautious on if the command returns the full data and/or not some other response in a corner case).

illume (changes requested):
Thanks for this. 

I left some questions.

Can you please update the commit messages?

plugins/aks-desktop/src/components/GitHubPipeline/hooks/useWorkloadIdentitySetup.ts
```diff
@@ -87,10 +87,15 @@ export const useWorkloadIdentitySetup = (): UseWorkloadIdentitySetupReturn => {
...
+      }
```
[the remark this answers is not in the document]
illume: I also made this point in another comment... but I think additionally the warnings might be logged and I'm not sure of their contents.

plugins/aks-desktop/src/utils/azure/identityRoles.ts
gambtho: that was backwards, thanks

gambtho: and yes, the rbac permissions aren't great -- seperate threads with that team to figure out how to get a better least privilege model for things like this

illume (changes requested):
Can you please update the commit messages to follow our formatting guidelines?

gambtho:
Consolidated into #598 per maintainer request. All review feedback has been addressed in the consolidated PR.

PR #564, opened 2026-04-02
feat: deterministic deploy workflow + K8s manifest generators

### LINKED ISSUE #552: Fast Path Pipeline: Deterministic deploy templates
Part of #557

## Problem

The Copilot agent generates a different workflow each time — and the generated workflows have recurring bugs: single-job design, missing permissions, unpinned kubelogin, fragile `sed`-based image substitution, and no explicit Dockerfile path. These are the root causes of most deploy failures.

When we already have the user's config (cluster, namespace, ACR, Dockerfile path, container settings), we can generate a correct workflow and manifests deterministically from tested templates.

## Changes

Create `fastPathTemplates.ts` with three pure functions:

- **`generateDeployWorkflow()`** — produces a two-job workflow (`buildImage` + `deploy`) based on the proven aks-devhub pattern. Fixes all known bugs: pinned `kubelogin-version: 'v0.1.6'`, `Azure/k8s-deploy@v5` for image substitution, `actions: read` permission, explicit `DOCKER_FILE` and `BUILD_CONTEXT_PATH` env vars, `continue-on-error` on annotations.
- **`generateDeploymentManifest()`** — Deployment YAML parameterized from ContainerConfig (replicas, resources, probes, security context, anti-affinity, pipeline annotations)
- **`generateServiceManifest()`** — Service YAML with configured type, ports, and selector

All functions are pure (config in → string out) with comprehensive test coverage.

gambtho:
## Summary

Pure-function template generators for the fast path pipeline — produces deploy-to-aks workflow YAML and K8s manifests deterministically from config, replacing non-deterministic agent generation:

- **`generateDeployWorkflow()`** — GitHub Actions workflow with two-job split (build + deploy), pinned kubelogin v0.1.6, `actions: read` permission, explicit Dockerfile path, `Azure/k8s-deploy@v5` for image substitution
- **`generateDeploymentManifest()`** — K8s Deployment with configurable replicas, resource requests/limits, optional probes and security context
- **`generateServiceManifest()`** — K8s Service with configurable type (ClusterIP/LoadBalancer), ports

## Changes Made

- New `fastPathTemplates.ts` — `PipelineConfig` and `ContainerConfig` types + three generator functions
- New `fastPathTemplates.test.ts` — comprehensive tests for parameterization, defaults, edge cases (nested Dockerfile paths, custom build context, probe configuration)

## Related Issues

Part of #557 (Fast Path Pipeline)
Closes #552

## Test Plan

- [x] Workflow includes correct permissions, pinned kubelogin, two-job structure
- [x] Workflow parameterizes Dockerfile path, build context, ACR, cluster, namespace
- [x] Deployment manifest respects replicas, resources, probes, security context
- [x] Service manifest supports ClusterIP and LoadBalancer types
- [x] Nested Dockerfile paths produce correct `DOCKER_FILE` and `BUILD_CONTEXT_PATH`
- [x] Type check passes

### DIFF
--- a/plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.test.ts
+++ b/plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.test.ts
@@ -43,7 +43,7 @@ describe('generateDeployWorkflow', () => {
 
   it('should use Azure/k8s-deploy@v5 for deployment', () => {
     const output = generateDeployWorkflow(baseConfig);
-    expect(output).toContain('Azure/k8s-deploy@v5');
+    expect(output).toContain('Azure/k8s-deploy@c8cfec839dc09896b3b8cc40cd13d04792680771'); // v5.1.0
     expect(output).not.toContain('kubectl apply');
   });
 
@@ -114,11 +114,11 @@ describe('generateDeployWorkflow', () => {
     expect(output).toContain('${{ secrets.AZURE_CLIENT_ID }}');
   });
 
-  it('should include concurrency block to cancel in-progress runs', () => {
+  it('should include concurrency block that queues rather than cancels in-progress runs', () => {
     const output = generateDeployWorkflow(baseConfig);
     const parsed = YAML.parse(output);
     expect(parsed.concurrency.group).toContain('github.workflow');
-    expect(parsed.concurrency['cancel-in-progress']).toBe(true);
+    expect(parsed.concurrency['cancel-in-progress']).toBe(false);
   });
 
   it('should annotate specific deployment, not all deployments in namespace', () => {
--- a/plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.test.ts
+++ b/plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.test.ts
@@ -137,6 +137,24 @@ describe('generateDeployWorkflow', () => {
     const output = generateDeployWorkflow(baseConfig);
     expect(output).toContain('"aks-project/pipeline-repo=${{ github.repository }}"');
   });
+
+  it('should throw when appName contains no alphanumeric characters', () => {
+    expect(() => generateDeployWorkflow({ ...baseConfig, appName: '???' })).toThrow(
+      /no alphanumeric characters/
+    );
+  });
+
+  it('should throw when appName is empty string', () => {
+    expect(() => generateDeployWorkflow({ ...baseConfig, appName: '' })).toThrow(
+      /no alphanumeric characters/
+    );
+  });
+
+  it('should throw when appName contains only special characters', () => {
+    expect(() => generateDeployWorkflow({ ...baseConfig, appName: '___###' })).toThrow(
+      /no alphanumeric characters/
+    );
+  });
 });
 
 const baseManifestConfig = {
@@ -334,6 +352,24 @@ describe('generateDeploymentManifest', () => {
     const container = parsed.spec.template.spec.containers[0];
     expect(container.readinessProbe.successThreshold).toBe(2);
   });
+
+  it('should throw when appName contains no alphanumeric characters', () => {
+    expect(() =>
+      generateDeploymentManifest({ ...baseManifestConfig, appName: '???' }, baseContainerConfig)
+    ).toThrow(/no alphanumeric characters/);
+  });
+
+  it('should throw when appName is empty string', () => {
+    expect(() =>
+      generateDeploymentManifest({ ...baseManifestConfig, appName: '' }, baseContainerConfig)
+    ).toThrow(/no alphanumeric characters/);
+  });
+
+  it('should throw when appName contains only special characters', () => {
+    expect(() =>
+      generateDeploymentManifest({ ...baseManifestConfig, appName: '___###' }, baseContainerConfig)
+    ).toThrow(/no alphanumeric characters/);
+  });
 });
 
 describe('generateServiceManifest', () => {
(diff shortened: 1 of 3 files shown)


plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,271 @@
...
+          periodSeconds: ${period}
+          timeoutSeconds: ${timeout}
+          failureThreshold: ${failure}
+          successThreshold: ${success}`;
+}
+
+/**
+ * Generates a Kubernetes Deployment manifest from container config.
+ */
+export function generateDeploymentManifest(config: ManifestConfig, cc: ContainerConfig): string {
```
sniok, line 201: could this function also generate yaml from plain js objects, similar to `generateYamlForContainer` in `yamlGenerator.ts`? concatenating strings to generate yaml is hard to maintain

also they seem very similar, maybe they can share some logic?

sniok (changes requested):

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache 2.0.
+
+import YAML from 'yaml';
+import type { ContainerConfig } from '../../DeployWizard/hooks/useContainerConfiguration';
+import { getProbeConfigs } from './probeHelpers';
+import { escapeYamlValue } from './yamlUtils';
+
+export interface WorkflowConfig {
+  appName: string;
```
illume, line 11: Can you please document these fields?

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,256 @@
...
+      actions: read
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    needs: [buildImage]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Azure login
+        uses: azure/login@v2
```
illume: It's better security practice to pin versions, rather than use a tag or the latest.

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,256 @@
...
+ */
+export function generateDeployWorkflow(config: WorkflowConfig): string {
+  const esc = escapeYamlValue;
+  return `name: Deploy to AKS
+
+on:
+  push:
+    branches: ["${esc(config.defaultBranch)}"]
+  workflow_dispatch:
+
```
illume: I think concurrency should not be the default for deploy.

Imagine two deploys are triggered. Which one should run in which order?


https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-concurrency-and-the-default-behavior

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

```

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,256 @@
...
+      initialDelaySeconds: probe.initialDelay,
+      periodSeconds: probe.period,
+      timeoutSeconds: probe.timeout,
+      failureThreshold: probe.failure,
+      successThreshold: probe.success,
+    };
+  }
+
+  // Security context
+  const securityContext: Record<string, boolean> = {};
```
illume: I'm not sure about this... are there UI toggles for this part?

I think we should default to the safer options, and allow people to opt out with a UI?

```ts
  const securityContext: Record<string, unknown> = {
    allowPrivilegeEscalation: false,
    runAsNonRoot: true,
    readOnlyRootFilesystem: true,
    capabilities: { drop: ['ALL'] },
    seccompProfile: { type: 'RuntimeDefault' },
  };
```

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,256 @@
...
+          kubectl annotate deployment --all \\
+            -n \${{ env.NAMESPACE }} \\
+            aks-project/pipeline-run-url=\${{ github.server_url }}/\${{ github.repository }}/actions/runs/\${{ github.run_id }} \\
+            "aks-project/pipeline-workflow=\${{ github.workflow }}" \\
+            --overwrite
+`;
+}
+
+export interface ManifestConfig {
+  appName: string;
```
illume, line 176: Please document this interface too?

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.test.ts
```diff
@@ -0,0 +1,293 @@
...
+describe('generateDeployWorkflow', () => {
+  it('should produce parseable workflow YAML with two jobs', () => {
+    const output = generateDeployWorkflow(baseConfig);
+    const parsed = YAML.parse(output);
+
+    expect(parsed.name).toBe('Deploy to AKS');
+    expect(Object.keys(parsed.jobs)).toContain('buildImage');
+    expect(Object.keys(parsed.jobs)).toContain('deploy');
+    expect(parsed.jobs.deploy.needs).toContain('buildImage');
+  });
```
illume, line 30: A test to make sure escaping doesn't break the github specific non-valid yaml would be good.

(untested)
```ts
it('should preserve GitHub Actions expressions (${{ }}) verbatim', () => {
  const output = generateDeployWorkflow(baseConfig);

  expect(output).toContain('${{ env.CLUSTER_NAME }}');
  expect(output).toContain('${{ github.sha }}');
});
```

plugins/aks-desktop/src/components/GitHubPipeline/utils/fastPathTemplates.ts
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * Generates a Kubernetes Deployment manifest from container config.
+ * Builds a plain JS object and serializes via YAML.stringify for correct
+ * indentation and escaping.
+ */
+export function generateDeploymentManifest(config: ManifestConfig, cc: ContainerConfig): string {
+  const container: Record<string, unknown> = {
+    name: config.appName,
+    image: `${config.acrName}.azurecr.io/${config.appName}:latest`,
```
illume: I'm wondering how this could be used to deploy other than latest?

Latest is kind of a security anti-pattern (but widely used I guess in some instances useful).

Can they enter a tag somewhere?

```yaml
image: `${config.acrName}.azurecr.io/${config.appName}:{{IMAGE_TAG}}`,
```

TODO: double check what k8s default is when image tag is absent (I think it defaults to latest)

illume (changes requested):
Thanks for this.

I left some notes.

sniok (approved):
lgtm, thanks

sniok:
oh sorry forgot about one thing, could you update the commit messages so they follow our guidelines? https://github.com/Azure/aks-desktop/blob/main/MAINTENANCE.md#git-commit-conventions

gambtho:
Consolidated into #598 per maintainer request. All review feedback has been addressed in the consolidated PR.

PR #595, opened 2026-04-14
CI: Run plugin tests only when code changes

### LINKED ISSUE #594: [BUG] CI: skip plugin tests when plugin code is unchanged
the CI workflow runs tests for all 3 plugins on every push and PR, regardless of which files were actually changed. we should detect which plugins were modified and ensure that only the tests for those run in the CI

skoeva:
These changes extend `ci.yml` with a `find_changed_plugins` pre-job so only the plugins affected by a PR are tested, rather than running all three on every push.

Fixes: #594 

### Summary
- Adds a `find_changed_plugin` job that uses `git diff origin/main...HEAD` to dynamically build a matrix of only the plugins with changed files
- All jobs (`test`, `lint`, `build`, `security`) consume the dynamic matrix and are skipped entirely if no plugins changed
- `translations` job runs only when `aks-desktop`, `ai-assistant`, or `Localize/` is touched
- If any file outside `plugins/` or `Localize/` changes (e.g. `.github/`, root `package.json`, shared scripts), all plugins run as a fallback
- Follows the same pattern used in headlamp-k8s/plugins

### Testing

- [ ] Open a PR touching only `plugins/aks-desktop` and verify only `aks-desktop` jobs run
- [ ] Open a PR touching only `plugins/ai-assistant` and verify only `ai-assistant` jobs run
- [ ] Open a PR touching only `plugins/insights-plugin` and verify only `insights-plugin` jobs run
- [ ] Open a PR touching only `Localize/**` and verify only the `translations` job runs
- [ ] Open a PR touching a root-level file (e.g. `package.json`) and verify all plugins run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,15 +11,85 @@ permissions:
 
 env:
   NODE_VERSION: "20"
-  CACHE_KEY_PREFIX: "node-modules"
 
 jobs:
+  find_changed_plugins:
+    name: Find changed plugins
+    runs-on: ubuntu-latest
+    outputs:
+      plugins: ${{ steps.detect.outputs.plugins }}
+      has_translation_plugins: ${{ steps.detect.outputs.has_translation_plugins }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Fetch base branch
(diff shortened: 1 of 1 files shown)


illume:
Can be done keeping the one file?

I remember doing something like this in the plugins repo with one file. Hrmm.

skoeva:
@illume I think so...I just thought it would be nice to split things up like in the main headlamp repo. I suppose one file would be good in the case that we add more plugins

sniok (approved):

PR #596, opened 2026-04-14
aks-desktop: Optimize vitest test config

skoeva:
These changes improve the vitest test setup for `aks-desktop` by enabling mock isolation, separating slow a11y tests, and adding coverage reporting.

### Summary

- Introduces `vitest.config.ts` to replace the inherited `headlamp-plugin` test config
- Enables `clearMocks: true` globally to prevent mock state from leaking between tests
- Splits tests into `unit` and `a11y` projects so guidepup a11y tests can be run independently
- Adds `@vitest/coverage-istanbul` and coverage configuration matching Headlamp's frontend setup
- Adds `test:unit`, `test:a11y`, and `test:coverage` npm scripts
- Removes 40 redundant `@vitest-environment jsdom` annotations across test files

### Testing

- [x] Run `cd plugins/aks-desktop && npm test` and ensure all tests pass
- [x] Run `npm run test:unit` and verify only unit tests run
- [x] Run `npm run test:a11y` and verify only guidepup tests run
- [x] Run `npm run test:coverage` and verify a coverage report is generated

### DIFF
--- a/plugins/aks-desktop/package-lock.json
+++ b/plugins/aks-desktop/package-lock.json
@@ -22,6 +22,7 @@
         "@testing-library/react": "^16.3.2",
         "@types/libsodium-wrappers": "^0.7.14",
         "@types/react": "^19.2.0",
+        "@vitest/coverage-istanbul": "^3.2.4",
         "eslint-plugin-no-barrel-files": "^1.2.2",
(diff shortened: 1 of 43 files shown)


illume (approved):
Thanks! 🎉 

Yeah, some nice improvements.

PR #609, opened 2026-04-21
aks-desktop: Extract shared EmptyStateCard component

### LINKED ISSUE #591: Move `EmptyStateCard` to `shared/` and use it in `LogsTab`
`EmptyStateCard` currently lives in `Metrics/components/` but is general purpose
`LogsTab` has an inline empty state that duplicates it
Move to `shared/`, update the `Metrics` import, and use it in `LogsTab`

skoeva:
These changes move `EmptyStateCard` out of `Metrics/components/` into `shared/` and replace the duplicate inline empty state in `LogsTab` with it.

Fixes: #591 

### Summary

- Moves `EmptyStateCard` and its props interface to `components/shared/`
- Updates the `MetricsTab` import path
- Replaces the inline `Card`/`Icon`/`Typography` block in `LogsTab` with `EmptyStateCard`

### Testing

- [x] Verify the empty state renders correctly in both `MetricsTab` and `LogsTab` when no deployments exist


sniok (approved):
nice, thanks!

PR #626, opened 2026-04-29
aksd: harden App Insights telemetry to anonymous app-usage counts

gambtho:
## Summary

Reduce App Insights telemetry shipped from the Headlamp frontend to anonymous app-usage counts only. An external review flagged the previous implementation as too liberal.

- **Lock down SDK config:** disable AJAX/fetch/exception/cookies/storage/auto-route auto-collection. Drop the unconditional `trackPageView()` call (URLs in Headlamp encode cluster, namespace, and resource names).
- **Allowlist event types** in the redux listener middleware. A future `HeadlampEventType` addition that accidentally encodes user data in its name string cannot leak — it must be opted into telemetry explicitly. `action.payload.data` is still never forwarded.
- **Replace `trackException`** (which shipped the full `Error` including message and stack) with `trackEvent('exception', { errorName: error.name })`. Ships only the constructor name (e.g. `TypeError`, `KubeApiError`) — no message, no stack.
- **Telemetry initializer** as defense-in-depth: strips `ai.user.*` / `ai.session.id` / `ai.location.ip` envelope tags and clears `uri` / `refUri` / `url` fields on every outgoing envelope, regardless of how they got there.

Net result: each outgoing telemetry envelope contains an allowlisted event name (`headlamp.list-view`, `headlamp.delete-resource`, …) and, for `'exception'`, `errorName`. No URL, user id, session id, IP, cluster, namespace, resource name, error message, or stack.

## Test plan

- [x] `npm run tsc` clean
- [x] `npm run lint` clean
- [x] New unit tests pass (13 total across `analytics.test.ts`, `headlampEventSlice.test.tsx`, `ErrorBoundary.test.tsx`)
- [ ] Manual smoke: run with `REACT_APP_APPINSIGHTS_CONNECTION_STRING` set; confirm in browser devtools (Network → `track`) that envelopes contain no `ai.user.*` / `ai.session.id` / `ai.location.ip` tags, no `uri` / `refUri` / `url`, no `PageView`, no `RemoteDependencyData`, and that triggering an error in an `ErrorBoundary` produces a single `EventData` envelope with `name: 'exception'` and `properties: { errorName: ... }` only.

### DIFF
--- a/frontend/src/components/common/ErrorBoundary/ErrorBoundary.test.tsx
+++ b/frontend/src/components/common/ErrorBoundary/ErrorBoundary.test.tsx
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render } from '@testing-library/react';
+import { Provider } from 'react-redux';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import * as analytics from '../../../lib/analytics';
+import store from '../../../redux/stores/store';
+import ErrorBoundary from './ErrorBoundary';
+
+class CustomDomainError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'CustomDomainError';
+  }
+}
+
+function Boom({ error }: { error: Error }): JSX.Element {
+  throw error;
+}
+
+describe('ErrorBoundary telemetry', () => {
+  let trackEventSpy: ReturnType<typeof vi.spyOn>;
+  let consoleErrorSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    trackEventSpy = vi.spyOn(analytics, 'trackEvent').mockImplementation(() => {});
+    // React logs caught errors via console.error in dev — silence to keep test output clean.
+    consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    trackEventSpy.mockRestore();
+    consoleErrorSpy.mockRestore();
+  });
+
+  it('calls trackEvent with exception name and only error.name in properties', () => {
+    const err = new CustomDomainError(
+      "namespace 'payments' not found at https://example.test/api/v1/namespaces/payments"
+    );
+
+    render(
+      <Provider store={store}>
+        <ErrorBoundary fallback={<div>fallback</div>}>
+          <Boom error={err} />
+        </ErrorBoundary>
+      </Provider>
+    );
+
+    expect(trackEventSpy).toHaveBeenCalledWith('exception', { errorName: 'CustomDomainError' });
+
+    // No call shipped the message or stack.
+    for (const call of trackEventSpy.mock.calls) {
+      const [, properties] = call;
+      if (properties) {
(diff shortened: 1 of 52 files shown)


skoeva (changes requested):
You could update the snapshots in the frontend since the CI seems to be failing

frontend/src/lib/analytics.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function makeEnvelope(overrides: Partial<ITelemetryItem> = {}): ITelemetryItem {
+  return {
+    name: 'Microsoft.ApplicationInsights.Event',
+    time: new Date().toISOString(),
+    ...overrides,
+  } as ITelemetryItem;
+}
+
+describe('privacyTelemetryInitializer', () => {
```
skoeva, line 30: this could be squashed into the commit introducing the function

frontend/src/redux/headlampEventSlice.test.tsx
```diff
@@ -111,16 +111,52 @@ describe('eventsSlice', () => {
...
 
-      expect(trackEventSpy).toHaveBeenCalledWith('test-event-type');
+      expect(trackEventSpy).toHaveBeenCalledWith('headlamp.list-view');
       expect(window.appInsights).toBeUndefined();
 
       window.appInsights = originalAppInsights;
       trackEventSpy.mockRestore();
     });
+
+    it('does not call trackEvent for an event type outside the allowlist', () => {
```
skoeva, line 126: these tests could also be squashed into the commit introducing the allowlist

frontend/src/components/common/ErrorBoundary/ErrorBoundary.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    super(message);
+    this.name = 'CustomDomainError';
+  }
+}
+
+function Boom({ error }: { error: Error }): JSX.Element {
+  throw error;
+}
+
+describe('ErrorBoundary telemetry', () => {
```
skoeva, line 35: these tests could be squashed into the last commit

frontend/src/lib/analyticsPrivacy.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Strip identifying tags and URL fields from every outgoing telemetry envelope.
+ *
+ * This runs on the SDK's send pipeline and is the catch-all that protects us
+ * if a future SDK upgrade re-enables auto-collection or if a new call site
+ * forgets the "no identifiers" rule. It is intentionally aggressive.
+ *
+ * Pure function — kept in a side-effect-free module so it can be unit-tested
+ * without dragging the AppInsights bootstrap into the test's module graph.
+ */
+export function privacyTelemetryInitializer(envelope: ITelemetryItem): void {
```
skoeva, line 32: maybe the old commit is redundant then? it can be dropped

skoeva (changes requested):
I think the changes look good, but they could probably all be squashed into a few commits. the updated tests/snapshots should not be isolated, and look at redundant commits

frontend/src/lib/analyticsPrivacy.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Mutates the provided `envelope` in place (as required by the
+ * `ITelemetryItem` initializer contract). Deterministic and free of external
+ * side effects — it touches only the envelope it is given, so it can be
+ * unit-tested without dragging the AppInsights bootstrap into the test's
+ * module graph.
+ */
+export function privacyTelemetryInitializer(envelope: ITelemetryItem): void {
+  envelope.tags = envelope.tags ?? {};
+  delete envelope.tags['ai.user.id'];
```
sniok, line 34: the tags are still present in the event, although we could probably keep them since they're not identifying users 

Image

frontend/src/components/common/ErrorBoundary/ErrorBoundary.test.tsx
gambtho: i think we should squash merge in general, but fair

frontend/src/lib/analyticsPrivacy.ts
gambtho: agreed -- going leave in case someone else disagrees

sniok (dismissed):

skoeva (approved):

Issue #634, opened 2026-05-04
[BUG] Info Tab shows blank screen


danielsollondon:
## Bug Description

<img width="1305" height="439" alt="Image" src="https://github.com/user-attachments/assets/198a9f25-c453-4768-9ae9-f117df9c630f" />

This should show quota etc.

AKS desktop Version
0.6.0
Headlamp Version
0.40.0

skoeva:
this is a two-part issue, pushing the fixes for this up

willspacecase:
Any updates on this? Does the fix cover both fresh load and reload scenarios?

PR #639, opened 2026-05-07
upstreamable: frontend: ProjectDetails: Fix stale custom tabs

skoeva:
These changes fix stale custom project detail tabs by scoping plugin tab state to the current route and resolved project identity.

Related: #634

### Summary

- Adds a project identity key from the route name, project id, clusters, and namespaces
- Remounts `ProjectDetailsContent` when the route or resolved project identity changes
- Stores custom/plugin tabs separately from default tabs and tags them with the current project key
- Clears project-specific plugin tabs while async `isEnabled` checks reload for the current project
- Ignores stale async tab eligibility results after the project changes
- Prevents blank tab content by resetting the selected tab when the previous tab no longer exists
- Keeps default project tabs visible while custom links are being reevaluated

### Testing

- [x] Manually verified navigating from a project with a custom Info tab to a project without that tab no longer leaves the stale Info tab visible
- [x] Manually verified hard reload on a project without an eligible info does not show stale tab content
- [x] Confirmed `frontend/src/components/project/ProjectDetails.tsx` has no TS errors

### DIFF
--- a/frontend/src/components/project/ProjectDetails.tsx
+++ b/frontend/src/components/project/ProjectDetails.tsx
@@ -87,6 +87,15 @@ const DEFAULT_TABS: Record<string, ProjectDetailsTab> = {
   },
 };
 
+function getProjectKey(project: ProjectDefinition, routeName?: string) {
+  return JSON.stringify({
+    routeName: routeName ?? project.id,
+    id: project.id,
+    clusters: [...project.clusters].sort(),
+    namespaces: [...project.namespaces].sort(),
+  });
+}
+
 export default function ProjectDetails() {
   const { t } = useTranslation();
   const { name } = useParams<ProjectDetailsParams>();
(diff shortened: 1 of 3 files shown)


tejhan (dismissed):
Lgtm. Outstanding build failure appears to be unrelated.

frontend/src/components/project/ProjectDetails.tsx
```diff
@@ -95,8 +104,9 @@ export default function ProjectDetails() {
   if (isProjectLoading || !project || !name) {
     return <Loader title={t('Loading')} />;
   }
-  // Key is provided to make sure we remount this component
-  return <ProjectDetailsContent key={name} project={project} />;
+  // Remount when the route or resolved project identity changes.
+  const projectKey = getProjectKey(project, name);
+  return <ProjectDetailsContent key={projectKey} project={project} projectKey={projectKey} />;
```
sniok: if we're passing projectKey as a key prop to this component then it will reset all internal state when it changes, so I don't think we actually need to pass it as projectKey

yolossn (changes requested):
@skoeva tested this and works as expected. Once the conflicts are fixed this can be merged

PR #660, opened 2026-05-12
upstreamable: frontend: Fix mobile navbar logo color

### LINKED ISSUE #638: [BUG] AKS desktop logo not responsive on mobile
Related: https://github.com/Azure/aks-desktop/issues/516

here's an example:

Image

### LINKED ISSUE #659: [BUG] upstreamable: Search icon not responsive on mobile
Image

related: #638

skoeva:
This change makes the mobile logo/sidebar control inherit the app bar foreground color so it remains readable with custom themes.

Fixes: #638

### Summary

- Updates the mobile logo/sidebar button to use `color="inherit"`
- Lets the button inherit the color already computed by the parent app bar from `theme.palette.navbar`
- Improves custom theme support for mobile app bar controls, including custom SVG logos that use `currentColor`
- Leaves GlobalSearch unchanged because equivalent behavior landed upstream in kubernetes-sigs/headlamp#4930

### Testing

- [x] Manually verified the mobile logo text inherits the navbar foreground color
- [x] `git diff --check upstream/headlamp-downstream...HEAD`
- [x] Focused HeadlampButton Storybook tests (5 passed)

Mobile navbar controls using the navbar foreground color


tejhan (commented):
Changes LGTM, although it looks like there are some potentially unrelated test run failures, might need a re-run / update.

illume (commented):
Is this change fixed in Headlamp? In that case maybe we don't need this PR anymore.

skoeva:
Rebuilt this branch on the current `headlamp-downstream` tip. The PR is now one commit changing only `HeadlampButton`.

I removed the `GlobalSearch` change because equivalent behavior landed upstream in kubernetes-sigs/headlamp#4930 and is already present on the base branch. I kept the logo/sidebar change because the current Headlamp implementation still forces `theme.palette.text.primary` instead of inheriting the AppBar foreground color.

tejhan (changes requested):
Fix LGTM, but this might have to be re-done via a patch from Headlamp now due to the restructuring. (These files are no longer in this repo).

The bug still exists for mobile, so this fix is good to have.

PR #698, opened 2026-06-01
insights-plugin: bump to v0.2.6

flyth:
[v0.2.6](https://github.com/inspektor-gadget/insights-plugin/releases/tag/v0.2.6) adds localization support and improved a11y coverage as well as two more gadgets to the Insights tab (CUDA Profile + CUDA Memory).

PR #712, opened 2026-06-02
insights-plugin: include locales dir

mqasimsarfraz:
https://github.com/Azure/aks-desktop/pull/698 missed extracting the localization files when downloading artifact for insights-plugin, this change fixes it!


flyth (approved):
Thanks!

PR #718, opened 2026-06-05
aksd: restore App Insights usage telemetry without PII

gambtho:
## Summary

Restores App Insights usage telemetry on the AKS Desktop plugin. The
plugin owns App Insights end-to-end — the Headlamp shell is intentionally
telemetry-inert.

This PR pairs with headlamp PR #723 (already merged into the submodule
at `0d6284c1a`), which removes the fork-side App Insights bootstrap and
the install-id / app-info IPC bridges.

## Architecture

**Structural chokepoint.** The `ApplicationInsights` SDK instance is
module-private inside `src/telemetry/index.ts`. There is no
`window.appInsights`, no exported handle. Five typed `track*` wrappers
(`trackSessionStart`, `trackClusterShape`, `trackFeature`,
`trackException`, `trackPluginsLoaded`) are the only path to
`ai.trackEvent`, and each hard-codes its envelope name and property
shape.

**Sync data sources.** Install ID lives in renderer `localStorage`
(`getOrCreateInstallId`, with an in-memory session cache so the UUID is
stable even when `setItem` throws). Host info is read from
`window.desktopApi.platform` (set by the headlamp preload bridge) plus
`navigator.userAgent` for arch and Electron version. No IPC, no new
fork env vars — runtime UA parsing reports the actual running arch on a
universal binary, more accurate than a build-time constant.

**Sanitized values only.** `src/telemetry/schema.ts` clamps every
property: k8s kinds to the built-in set or `CustomResource`, regions via
shape regex, AKS tiers to `Free|Standard|Premium|Unknown`, event
statuses to a closed enum, locales to language code, counts to buckets
(node/namespace bucketers return an explicit `'0'` for zero/negative
inputs rather than mislabelling them into the lowest range).
`headlamp.cluster-shape` only fires when every field is non-null —
better to lose the signal than mix `Unknown` with real values.

**Anonymous correlation.** A minimal `addTelemetryInitializer` stamps
`ai.user.id` with the install UUID on every envelope so the Azure
Portal Users metric correlates by install rather than by SDK session.
The install id lives ONLY as the `ai.user.id` envelope tag — never as
a regular property dimension that would be queryable / exportable as
data.

**Honest payload extraction.** `extractKindFromPayload` in
`src/telemetry/setup.ts` returns `undefined` (not `'Unknown'`) when an
event's `resource.kind` / `resourceKind` is absent, and returns
`undefined` for plural-resource batches when any entry lacks a kind.
Avoids emitting a misleading bucket when the source data isn't there.

**Plugin-only exception capture.** `TelemetryErrorBoundary` (class
component, since `componentDidCatch` has no hooks equivalent) wraps
each `registerRoute` and `registerProjectDetailsTab` in the AKS plugin
and forwards `Error.name` through `trackException`. The boundary itself
is registered unconditionally (routes and tabs are available in both
app and browser mode); `TelemetryBoot` — which initialises App Insights
— is registered only inside `isRunningAsApp`, so no envelopes are ever
posted from a browser host. Project overview sections and header actions
are intentionally NOT wrapped — spec scopes boundary coverage to routes
and tabs.

## Privacy posture

- **Opt-out toggle, default on.** Settings panel
  (`registerPluginSettings('aks-desktop', TelemetrySettings)`) adds a
  "Send anonymous usage data" toggle inside the existing AKS Desktop
  settings section alongside Preview Features. Takes effect on next
  launch (per spec, not live opt-out). Backed by a dedicated
  `ConfigStore` at `'aks-desktop.telemetry'`.
- **Closed envelope-name set.** Five hard-coded literals in
  `index.ts`: `headlamp.session-start`, `headlamp.cluster-shape`,
  `headlamp.feature`, `headlamp.exception`, `headlamp.plugins-loaded`.
- **Closed feature-name allowlist.** `KNOWN_FEATURE_TYPES` /
  `TELEMETRY_EVENT_ALLOWLIST` gate which redux event types route to
  `headlamp.feature`. A future upstream `HeadlampEventType` change
  can't smuggle data through the `feature` property.
- **Closed event-status enum.** Clamped via `sanitizeStatus`.
- **SDK auto-collection disabled.** `disableFetchTracking`,
  `disableAjaxTracking`, `disableExceptionTracking`,
  `disableCookiesUsage`, `isStorageUseDisabled`,
  `enableAutoRouteTracking: false`. Combined with the chokepoint, this
  replaces the defense-in-depth allowlists from earlier drafts.
- **Anonymous install ID.** UUID v4 in `localStorage` (opaque `InstallId`
  brand type prevents accidental PII injection at callsites), never
  concatenated with PII. Reset by clearing site data.

## File layout

```
src/telemetry/
  index.ts         module-private AI; initTelemetry + 5 track* helpers
  setup.ts         redux-callback registration + extractKindFromPayload
  schema.ts        sanitizers + closed vocabularies + count bucketers
  installId.ts     localStorage UUID + session cache
  appInfo.ts       desktopApi.platform + navigator.userAgent parser
  *.test.ts        unit suites for each of the above

src/components/
  TelemetryBoot.tsx              mounts via registerAppBarAction (app-only)
  TelemetryErrorBoundary.tsx     class component, wraps routes + tabs
  PluginSettings/
    TelemetrySettings.tsx          opt-out toggle, default on
    telemetrySettingsStore.ts      ConfigStore('aks-desktop.telemetry')
    PreviewFeaturesSettings.tsx    pre-existing, untouched
    previewFeaturesStore.ts        pre-existing, untouched
```

The production App Insights connection string is hardcoded as a
constant inside `TelemetryBoot.tsx` (connection strings are
addresses, not credentials). Override at build time by exporting
`REACT_APP_APPINSIGHTS_CONNECTION_STRING` — headlamp-plugin's Vite
config substitutes that env var into `import.meta.env` at bundle
time and the override wins over the default.

## Commits

In branch order (oldest first):

1. `deps: add App Insights, i18n strings, submodule bump`
2. `settings: add TelemetrySettings and its ConfigStore-backed opt-out`
3. `aks: surface cluster location, tier, kubernetesVersion in capabilities`
4. `telemetry: add private-AI chokepoint with sanitizers and typed helpers`
5. `telemetry: boot on mount, wrap plugin subtree, emit cluster-shape`

Each commit builds and lints independently.

## Test plan

- [x] `npm run tsc` — clean
- [x] `npm run lint` — clean
- [x] `npm run format -- --check` — clean
- [x] `npm test` — CI `Test (aks-desktop)` job passes
- [x] CI `Translations` job passes against the bumped submodule
- [ ] Manual: launch app, confirm Telemetry settings panel renders with
      toggle inside the AKS Desktop settings section
- [ ] Manual: opt out, restart, confirm no envelopes post (no
      `window.appInsights` exists by design — verify via network panel)
- [ ] Manual: opt in, navigate to an AKS cluster, confirm
      `headlamp.cluster-shape` fires once per resource id with all
      sanitized properties non-null
- [ ] Manual: confirm `localStorage.getItem('aksdInstallId')` returns
      a UUID v4 after first interaction, reload preserves it, DevTools
      "Clear Site Data" mints a new one
- [ ] Manual: trigger a route-tree error and confirm the Alert
      fallback renders and a `headlamp.exception` envelope posts with
      `errorName: 'Error'`

## AI assistance disclosure

Implementation and PR description authored with assistance from Claude
(Anthropic). All commits manually reviewed and tested locally before
push.


plugins/aks-desktop/src/telemetry/schema.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export type AksTier = 'Free' | 'Standard' | 'Premium' | 'Unknown';
+const VALID_AKS_TIERS: ReadonlySet<string> = new Set(['Free', 'Standard', 'Premium']);
+
+export function sanitizeTier(tier: string | null | undefined): AksTier {
+  return tier && VALID_AKS_TIERS.has(tier) ? (tier as AksTier) : 'Unknown';
+}
+
+export type NodeCountBucket = '1-5' | '6-20' | '21-100' | '100+';
+export function bucketNodeCount(n: number): NodeCountBucket {
```
sniok, line 133: these kind of histograms are usually used after the fact from raw values when analyzing metrics
are we sure we want to do this client side? unless this was done for privacy reasons

plugins/aks-desktop/src/telemetry/index.ts
```diff
@@ -0,0 +1,198 @@
...
+  sanitizeRegion,
+  sanitizeStatus,
+  sanitizeTier,
+} from './schema';
+import { registerReduxCallback } from './setup';
+
+// Module-private SDK handle. Only the typed track* functions below can
+// post envelopes.
+let ai: ApplicationInsights | null = null;
+let hasInitialized = false;
```
sniok: I think `hasInitilized` is a bit redundant since we can just check `if (ai === null)`

plugins/aks-desktop/src/components/PluginSettings/TelemetrySettings.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const telemetrySettingsStore = panel.store;
+export const isTelemetryEnabled = panel.isEnabled;
+export const useTelemetryEnabled = panel.useEnabled;
+
+// Register strings with i18next-parser. The factory looks them up at
+// render time via t(opts.titleKey); these stub calls let the static
+// extractor see them.
+// eslint-disable-next-line no-unused-vars
+function _i18nKeys(t: (k: string) => string) {
```
sniok: this looks so sketchy 😄 

are we sure we need a generic boolean panel renderer? maybe we can just make a specific component for telemetry settings so we don't need to do magic with i18n like this

plugins/aks-desktop/src/components/ClusterCapabilityCard/ClusterCapabilityCard.tsx
```diff
@@ -46,7 +56,29 @@ function ClusterCapabilityCard({ project }: ClusterCapabilityCardProps) {
...
+  // Live k8s counts feed the cluster-shape envelope; null-guard lives
+  // inside trackClusterShape.
+  const [nodes] = K8s.ResourceClasses.Node.useList({ cluster });
+  const [namespaces] = K8s.ResourceClasses.Namespace.useList({ cluster });
+
+  useEffect(() => {
+    if (!capabilities || !subscription || !resourceGroup || !cluster) return;
+    if (!nodes || !namespaces) return;
+    const azureResourceId = `/subscriptions/${subscription}/resourceGroups/${resourceGroup}/providers/Microsoft.ContainerService/managedClusters/${cluster}`;
+    if (emittedShapeFor.has(azureResourceId)) return;
```
sniok: maybe this check can be moved to `truckClusterShape` that can keep track of which clusters were already tracked, it might make this useEffect a bit simpler

plugins/aks-desktop/src/components/PluginSettings/TelemetrySettings.tsx
sniok: or these strings could be passed directly into createBooleanSettingPanel as a ReactNode

```tsx
import { Trans } from 'react-i18next';

const panel = createBooleanSettingPanel({
  title: <Trans>Telemetry</Trans>
...
```

and then used directly like so 

```tsx
<Typography variant="h6">{ops.title}</Typography>
```

plugins/aks-desktop/src/telemetry/index.ts
gambtho: naming was bad, useful to prevent retry

plugins/aks-desktop/src/telemetry/schema.ts
gambtho: that was the intent -- but i could go either way on it?

sniok (dismissed):
thanks!

sniok (approved):
thanks!

PR #722, opened 2026-06-09
aks-desktop: Clean up knip findings

skoeva:
These changes resolve all outstanding knip findings by tuning the knip config for the current plugin layout, unexporting internal-only symbols, and removing dead helpers left behind when the ai-assistant plugin moved upstream.

### Testing

- [x] `npm run knip` reports zero issues
- [x] No TS errors in any of the touched files


illume (approved):
🎉 thanks!

PR #734, opened 2026-06-18
Strip upstream CORS headers in proxy ModifyResponse

yolossn:
## Description

Strip CORS headers from upstream responses. Some upstreams (notably the K8s API server behind `az connectedk8s proxy` / arcProxy, which sets `Access-Control-Allow-Origin: *`) attach their own CORS headers. In dev mode Headlamp wraps the router in handlers.CORS, which would then emit a second `Access-Control-Allow-Origin` value and the app rejects the response.


backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -420,6 +420,24 @@ func (c *Context) SetupProxy() error {
 
 	proxy := httputil.NewSingleHostReverseProxy(URL)
 
+	// Strip CORS headers from upstream responses. Some upstreams (notably the
+	// K8s API server behind `az connectedk8s proxy` / arcProxy, which sets
+	// `Access-Control-Allow-Origin: *`) attach their own CORS headers. In dev
+	// mode Headlamp wraps the router in handlers.CORS, which would then emit a
+	// second `Access-Control-Allow-Origin` value and the app rejects the
+	// response.
+	proxy.ModifyResponse = func(resp *http.Response) error {
```
sniok, line 429: since we only add cors headers in the dev mode, should we also only remove incoming headers in the dev mode as well?

PR #774, opened 2026-07-17
aksd: telemetry: stamp app version on every event

gambtho:
## Problem

The recent telemetry privacy hardening only attached the `appVersion` property to `headlamp.session-start` and `headlamp.exception` events. Every other event — `headlamp.feature`, `headlamp.cluster-shape`, `headlamp.plugins-loaded` — carried no app version at all, so those events could not be correlated by release in the `appVersion` custom dimension our queries read.

Observed in production: `headlamp.feature` envelopes arrive with no version field.

## Change

Generalize the `appVersion` stamping in `send()` from `headlamp.exception` to all events, so every envelope carries the version as a `customDimensions.appVersion` property. `appVersion` is already in the `TELEMETRY_PROPERTY_KEYS` allowlist, so no schema change is needed.

## Testing

- `envelope.test.ts` now asserts the serialized `headlamp.feature` batch carries `properties.appVersion` (end-to-end serialization test).
- All 186 telemetry tests pass; `tsc`, `lint`, and `format --check` clean under Node 20.

### DIFF
--- a/plugins/aks-desktop/src/telemetry/envelope.test.ts
+++ b/plugins/aks-desktop/src/telemetry/envelope.test.ts
@@ -65,6 +65,7 @@ describe('real Application Insights envelope', () => {
     expect(featureEnvelope.data.baseData.properties).toEqual({
       feature: 'headlamp.logs',
       status: 'opened',
+      appVersion: '1.0.0-test',
     });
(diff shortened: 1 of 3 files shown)


illume (approved):
🎉 thanks!

PR #785, opened 2026-07-22
aksd: expand telemetry coverage for core workflows

gambtho:
## Description

Extends AKS-specific telemetry from three tracked workflows (project-create,
project-import, deploy) to eight, adding auth-login, auth-logout, cluster-add,
namespace-create, and project-delete.

Emitting accurate lifecycle events required fixing latent races in the auth and
namespace flows — in-flight polls and timers could resolve after unmount or after
the user cancelled, which would have produced both wrong telemetry and
setState-after-unmount warnings. Those fixes are included here.

## Type of Change

## Changes Made

**Telemetry infrastructure**

- `telemetry/schema.ts`: new `AKS_FEATURE_TYPES` array as the single source of
  truth for AKS feature names; `KNOWN_FEATURE_TYPES` now spreads it. Adds five
  error areas.
- `telemetry/aksFeature.ts` (new): `trackAksFeature(feature, status)` — a typed
  wrapper that swallows transport failures so telemetry can never break the
  workflow it measures.
- `hooks/useTelemetryFeatureOpened.ts` (new): emits `opened` on mount.

**Instrumented surfaces**

- Azure login (`AzureLoginPage`) — opened/started/succeeded/failed/cancelled,
  with `TimeoutError` distinguished from `UnknownError`.
- Azure logout (`useAzureProfilePage`) — started/succeeded/failed.
- Cluster registration (`RegisterAKSClusterDialog` + `Page`) — the dialog reports
  start/finish to the page via new `onRegistrationStarted`/`onRegistrationFinished`
  props so the page can attribute cancellation. As with `auth-login`, `cancelled`
  is reserved for an attempt the user actually started; opening the page and
  leaving reports only `opened`.
- Namespace creation (`CreateNamespace`) — full lifecycle, guarded by an attempt
  state machine that ignores stale async results. `cancelled` is emitted from the
  explicit Cancel action; cancellation on header-back navigation is deferred to a
  follow-up, since it requires replacing `SectionBox`'s `backLink` prop (which
  accepts only a route and exposes no click hook) with an explicit Back button.
- Project deletion (`AKSProjectDeleteButton`, `useProjectDeletion`) — opened and
  cancelled on the button, started/succeeded/failed around the deletion itself.

**Lifecycle fixes made to support the above**

- `AzureLoginPage`: polling interval moved from state to a ref; added mounted and
  attempt-generation guards; redirect timeout now cancelled on unmount.
- `useAzureProfilePage`: same guards around the logout redirect.
- `CreateNamespace`: added a double-submit guard and success-timer cleanup.
- `RegisterAKSClusterPage`: navigation timer is now cleared on unmount and can no
  longer be scheduled twice.

**Behavior change to note:** in `RegisterAKSClusterDialog`, the try/catch now
wraps only the `registerAKSCluster` call. Failures in the post-success path
(capability lookup, success messaging) no longer surface as registration
failures — which is the intent, but it is a change from previous behavior.

## Testing

- [x] Unit tests pass
- [x] Manual testing completed

New test files cover each instrumented surface, including the cancellation and
unmount paths that previously had no coverage.

### Test Cases

1. Each of the five new features emits its expected event sequence for the
   success, failure, and cancellation paths.
2. Unmounting mid-flight emits no terminal event and leaves no timers running.
3. A stale async result from a superseded attempt is ignored.
4. `trackAksFeature` does not propagate a throwing transport.
5. `AKS_FEATURE_TYPES` and `KNOWN_FEATURE_TYPES` stay in sync (vocabulary test).

## Reviewer Notes

- `useProjectDeletion.ts` shows a large line count but is almost entirely
  reindentation from the added try/catch. The behavioral delta is the three
  telemetry calls.
- `errorClass` is only narrowed where the code has a definite signal: the
  `auth-login` poll timeout reports `TimeoutError`. Every other new `trackError`
  call reports `UnknownError` rather than inferring a class from error text,
  which would risk leaking message content into a categorical field.
- The lifecycle/race hardening in `AzureLoginPage`, `useAzureProfilePage`, and
  `CreateNamespace` is independently reviewable and may be split into its own PR
  if reviewers prefer a smaller change.
- Namespace back-navigation cancellation is deliberately absent here and lands in
  a follow-up (`telemetry-namespace-back-cancel`), because it changes the header
  affordance from `SectionBox`'s built-in back link to an explicit button.


plugins/aks-desktop/src/components/AKS/RegisterAKSClusterPage.tsx
```diff
@@ -12,24 +14,67 @@ import RegisterAKSClusterDialog from './RegisterAKSClusterDialog';
 export default function RegisterAKSClusterPage() {
   const [open, setOpen] = useState(true);
   const history = useHistory();
+  // Starts 'idle' so that merely opening the page and leaving reports only
+  // `opened`. `cancelled` is reserved for a registration the user actually
+  // started and then backed out of, matching auth-login.
+  const terminalStatusRef = useRef<'idle' | 'active' | 'cancelled' | 'failed' | 'succeeded'>(
```
sniok, line 20: took me a bit to realize that this meant 'final status' and not that there's some kind of terminal (like xterm) in this component. maybe `finalStatusRef` would be easier to understand?

plugins/aks-desktop/src/components/CreateNamespace/CreateNamespace.tsx
```diff
@@ -45,6 +48,48 @@ function getStepLabel(t: (key: string) => string, step: NamespaceStepName): stri
...
+  finish: (
+    generation: number,
+    outcome: Extract<NamespaceCreationAttemptOutcome, 'failed' | 'succeeded'>
+  ) => boolean;
+  invalidate: () => void;
+  is: (generation: number, outcome: NamespaceCreationAttemptOutcome) => boolean;
+  start: () => number;
+}
+
+function createNamespaceCreationAttemptState(): NamespaceCreationAttemptState {
```
sniok, line 64: this could probably be a hook that could be reused for other feature tracking

sniok (approved):
thanks, left a couple of minor comments but overall it looks good

plugins/aks-desktop/src/components/CreateNamespace/CreateNamespace.tsx
gambtho: agreed -- but fixing in upcoming pr (same for naming)

PR #791, opened 2026-07-24
aks-desktop: Stop requiring ManagedNamespacePreview

### LINKED ISSUE #788: [BUG] Unable to create new project using Azure Managed Namespace
## Bug Description

The New Project UI shows an error message stating the `ManagedNamespacePreview` feature flag is required but the Azure Managed Namespace went into generally available status in November 2025; thus this flag should not be required.

## Steps to Reproduce

1. Go to Projects
2. Click on Create Project
3. See error

## Expected Behavior

Should not ask for feature flag to be enabled

## Actual Behavior

Unable to create project because this feature is not registered in my subscription.

## Screenshots

Image

## Environment

- **Plugin Version**: 0.9.0
- **Headlamp Version**: 0.43.0
- **Kubernetes Version**: 1.35.5
- **Azure CLI Version**: 2.88.0
- **Browser**: 
- **OS**: macOS

## Logs

If applicable, please include relevant logs:

```
Paste logs here
```

## Checklist

- [ ] I have searched for existing issues
- [ ] I have provided all necessary information
- [ ] I have tested with the latest version
- [ ] I have included steps to reproduce

skoeva:
AKS Managed Namespaces are generally available and no longer require the `ManagedNamespacePreview` subscription feature. The Create Project wizard still checked this feature and blocked users when it was unavailable. This removes the obsolete requirement so project creation can proceed without feature registration.

Fixes: #788

### Changes

- Remove the `ManagedNamespacePreview` warning and registration action from the Create Project wizard
- Stop checking feature registration when a subscription is selected
- Remove feature registration from Basics-step validation
- Remove the unused feature-check hook, Azure CLI registration helpers, types, stories, and test fixtures
- Update the Azure utilities documentation to remove the obsolete feature-registration responsibility

### Testing
- [x] `npm run test:unit -- src/components/CreateAKSProject` passes (123 tests)
- [x] Create a project in a subscription without `ManagedNamespacePreview` registered and confirm the wizard proceeds


sniok:
In the bundled azure cli we also install `"aks-preview"` extension (packge.json:57), could you check if it is still needed since managed namespaces are no longer in preview? Although it might be required by some other features we use

skoeva:
Thanks, I checked this. We still need `aks-preview` for the current bundled CLI: AKS Desktop pins Azure CLI `2.78.0`, while the core `az aks namespace` command group was added in Azure CLI `2.80.0` (Azure/azure-cli#32387), which is also the minimum version in the GA managed namespaces documentation. On `2.78.0`, removing the extension would break the namespace add/update/show/list/delete/get-credentials commands used by the project flow. I also checked our Azure command usage and found no other explicit dependency on `aks-preview`. I suggest keeping it in this PR, then removing the extension checks, UI, docs, and bundle entry together when we upgrade the bundled CLI to `2.80.0+`.

gambtho:
`Microsoft.ContainerService` registration is a separate requirement from the `ManagedNamespacePreview` flag — the provider has to be registered on a subscription to create managed namespaces there, GA or not. Any subscription with an AKS cluster has it already, and the wizard requires picking an existing cluster.

#860 keeps the helper and moves the call to the effect that watches the selected subscription, passing `--subscription` (it previously used the CLI default, which may not be the subscription the user picked). If this merges first, #860 restores it on rebase.

sniok (approved):
nice, thanks

PR #824, opened 2026-08-03
Use AKS desktop config dir for plugins and kubeconfigs

### LINKED ISSUE #766: aks desktop should point to its own plugin folder rather than headlamp folder
make sure aks desktop is using aks desktop folders for plugins (not headlamp folder). I mean the user config folders, currently aks desktop looks in the headlamp user folder for plugins. It should look in its own folder.

yolossn:
## What

Root the default plugins, user-plugins, and kubeconfigs directories at an
`AKS desktop` folder instead of the hardcoded `Headlamp` folder, in both the
Go backend and the Electron app.

## Why

These directories were rooted at `Headlamp`, so AKS Desktop shared them with a
co-installed Headlamp. Using `AKS desktop` (the app's userData folder name)
keeps them specific to AKS Desktop and alongside the app's other config.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] CI/CD changes
- [ ] Other: \***\*\_\_\_\*\***

### DIFF
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -73,7 +73,7 @@ jobs:
           echo "coverage=$testcoverage" >> $GITHUB_ENV
           echo "testcoverage_full_base64=$testcoverage_full_base64" >> $GITHUB_ENV
           echo "cleaning up..."
-          rm ~/.config/Headlamp/kubeconfigs/config
+          rm -f "$HOME/.config/AKS desktop/kubeconfigs/config"
         shell: bash
 
       - name: Run fuzz tests
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -913,6 +913,16 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
     // Directory doesn't exist or is not readable — ignore and continue.
   }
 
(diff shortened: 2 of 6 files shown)


skoeva (dismissed):
lgtm, thanks!

illume (changes requested):
This should be fixed inside Headlamp, such that it's configurable.

I'll take care of this.

yolossn:
Thanks @illume. I've updated the PR so plugins/user-plugins now use the existing --plugins-dir/--user-plugins-dir config (passed from the app) — no backend change for that.

The only remaining backend bit is kubeconfigs, since `MakeHeadlampKubeConfigsDir()` is hardcoded and has no config option.

illume:
Hey hey. 

Thanks @yolossn Can you please add this PR as a Headlamp repo one instead? 
 
headlamp-downstream is going away, and it means one less commit to upstream.
We'll release Headlamp before the next AKS desktop release.

(I also offer to do it if you prefer, let me know?)

illume (approved):
🎉 thanks!

PR #843, opened 2026-08-04
feat: add Contact us button to the app bar

### LINKED ISSUE #679: [FEATURE] feedback / contact us link in tool
## Feature Summary

We would like to make it easier for customers to engage with us, from the tool it is not obvious how customers can reach out and where.

Ask, can we add a contact us button that redirects to gh issues page to a pinned issue that shows how to engage us e.g. link to discord, when to raise an issue et. This should be accessible from most screens.

gambtho:
## Description

Adds a **Contact us** button to the top app bar, and repoints the Help menu's
"Open Issue" entry at the same destination, so there is one front door for
reaching the AKS Desktop team instead of two.

Both entry points open `https://aka.ms/aks-desktop/feedback`, a redirect to the
AKS Desktop feedback form. Keeping the destination behind a redirect means it
can be retargeted without cutting a release, and older builds keep working.

## Type of Change

## Related Issues

Closes #679

The `ErrorPage` crash button raised in review is split into #881 — it is a
submodule change, so it needs its own PR.

## Changes Made

- `utils/constants/contactUs.ts` — new `CONTACT_US_URL` constant, single source
  of truth for both entry points.
- `components/ContactUs/ContactUsButton.tsx` — app-bar action built on
  Headlamp's `ActionButton`, using the default `'action'` button style at every
  width to match Headlamp's own app-bar actions (see `SettingsButton`).
- `index.tsx` — registers the action as `{ id: 'aksd.contact-us', action }` so
  app-bar processors can address it, and replaces the `original-open-issue`
  Help-menu entry.
- `telemetry/schema.ts` — allowlists `aksd.feedback`. The telemetry layer
  silently drops unknown feature names, so an unlisted value would have made
  the event vanish with no error.
- `Localize/locales/en/plugin-translation.json` — the two new strings.

No new dependencies.

## Testing

- [x] Unit tests pass
- [x] Accessibility tested — see test cases 5 and 6
- [x] Link destination verified — `https://aka.ms/aks-desktop/feedback` returns
      `301` → `https://forms.cloud.microsoft/r/1jc9BUXTDb`

### Test Cases

1. Renders a button labelled "Contact us".
2. Clicking it calls `openExternalUrl` with `CONTACT_US_URL`.
3. Clicking it emits `trackFeature({ feature: 'aksd.feedback', status: 'opened' })`.
4. Navigation is attempted *before* the telemetry event is recorded, so the
   metric reflects a click that was actually acted on.
5. Rendered inside the `MenuItem` wrapper `TopBar` applies below the `sm`
   breakpoint, it produces no nested `MenuItem` — no `<li>` inside `<li>`.
6. In that same wrapper it stays keyboard-focusable and exposes an accessible
   name via `aria-label`.
7. `CONTACT_US_URL` survives `openExternalUrl`'s real validation — the helper
   returns silently on a malformed URL, so a typo would otherwise ship as a
   dead button with no error.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes

## Performance Impact

- [x] No performance impact

## Documentation Updates

- [x] No documentation updates needed — the user-facing guidance lives behind
      the redirect, deliberately, so it can change without a release.

## Reviewer Notes

Focus areas:

1. **The Help-menu change.** This replaces "Open Issue" rather than adding
   alongside it. Sending both to the same place was a deliberate call — worth
   confirming you agree.
2. **The app-bar button style.** Below the `sm` breakpoint `TopBar` moves
   app-bar actions into an overflow menu and wraps each in its own `MenuItem`.
   An earlier revision switched to `buttonStyle="menu"` at that width for a
   visible text label, but that nests a `MenuItem` inside a `MenuItem`: invalid
   `<li>` in `<li>`, and because the inner one is not focusable and the outer
   one carries no handler, the action could not be activated by keyboard at
   all. This follows the platform pattern instead — the `IconButton` stays
   focusable and its `aria-label` supplies the name.
3. **Known gap:** the Help-menu label `'Contact us'` is hardcoded English while
   the app-bar button uses `t('Contact us')`. `Headlamp.setAppMenu` runs at
   module scope with no translation function available, and the adjacent
   `'Documentation'` entry has the same limitation. Flagging rather than hiding.
4. The app-bar action is intentionally not wrapped in `TelemetryErrorBoundary`:
   Headlamp already wraps every app-bar action in its own `ErrorBoundary`, and
   this one's fallback is a visible MUI `Alert`, which in an app bar would be
   worse than a missing icon. Neither boundary covers the `onClick` path —
   React error boundaries do not catch throws from event handlers.


yolossn:
Ideally the same text/url should be used in the ErrorPage component too
[headlamp/frontend/src/components/common/ErrorPage/ErrorPage.tsx](https://github.com/Azure/aks-desktop/blob/4d00ea845c8f4faf2c7fde887f6a4bf9da2000c6/frontend/src/components/common/ErrorPage/ErrorPage.tsx)

illume (changes requested):
Please see the note by @yolossn ?

> Ideally the same text/url should be used in the ErrorPage component too [headlamp/frontend/src/components/common/ErrorPage/ErrorPage.tsx](https://github.com/Azure/aks-desktop/blob/4d00ea845c8f4faf2c7fde887f6a4bf9da2000c6/frontend/src/components/common/ErrorPage/ErrorPage.tsx)

pauldotyu:
@gambtho, the shortlink https://aka.ms/aks-desktop/feedback has been created

gambtho:
Yep, ErrorPage still files against `kubernetes-sigs/headlamp`. Split it into #881 — it's a submodule change so it needs its own PR.

Flagged there: the crash button's whole point is the prefilled stack trace, and the aka.ms link now lands on a Forms page that can't take one. So #881 keeps the structured `issues/new` URL and just fixes the repo + label. Open to going the other way if you'd prefer one destination.

Also in this push: the overflow-menu button wasn't keyboard-activatable at all (nested `MenuItem`, Copilot caught it), and the locale commit is trimmed down — 28 files → 9.

illume (approved):
🎉 thanks!

PR #849, opened 2026-08-06
upstreamable: frontend: Add Headlamp events for project, settings and plugin views

ashu8912:
## What

Adds `HeadlampEvent`s for views that previously dispatched nothing, so plugins can observe them:

| Event type | Payload |
| --- | --- |
| `PROJECT_LIST_VIEW` | `{ projects }` |
| `PROJECT_DETAILS_VIEW` | `{ project, resources }` |
| `PROJECT_DETAILS_TAB_CHANGE` | `{ project, tab, previousTab, resources }` |
| `CREATE_PROJECT` | `{ project, status }` |
| `DELETE_PROJECT` | `{ project, deleteNamespaces, status }` |
| `SETTINGS_VIEW` | `{ theme }` |
| `CLUSTER_SETTINGS_VIEW` | `{ cluster }` |
| `PLUGIN_LIST_VIEW` | `{ plugins }` |
| `PLUGIN_DETAILS_VIEW` | `{ plugin }` |

Each event gets a payload interface, a `useEventCallback` overload, a dispatch case in `headlampEventSlice`, and a re-export from the plugin registry so plugins can type against it. The `pluginLib` snapshot is updated accordingly.

## Notes

- `PROJECT_DETAILS_TAB_CHANGE` fires on the `selectedTab` value rather than the click handler, so programmatic tab switches (e.g. plugin-registered header actions) are captured too. The initially selected tab is not reported, since `PROJECT_DETAILS_VIEW` already covers it.
- `previousTab` is the full `ProjectDetailsTab` rather than just an id, matching `tab`.

## Testing

- `npm run tsc` — clean
- `npx vitest run src/redux/headlampEventSlice.test.tsx src/plugin/pluginLib.test.ts` — passing


illume:
Can you please move this to Headlamp repo instead? Headlamp will be released before AKS Desktop is released.

skoeva (commented):
this was merged upstream so I think it can safely be closed here

PR #883, opened 2026-08-18
plugins: Show headlamp org plugins when AKS optimized filter is off

yolossn:
## Description

When the AKS Desktop optimized toggle is disabled, show only plugins from the headlamp org (headlamp-k8s/plugins) instead of the full ArtifactHub catalog. These are known-safe plugins from Headlamp maintainers, making the disabled state a safer fallback.

Before:
image

After:
image


illume (approved):
🎉 thanks!

Issue #852, opened 2026-08-07
[BUG] Deleting resources from project view


pauldotyu:
## Bug Description

When deleting resources from within a project as soon as I click to delete a resource, the resource panel is kept open, and I get taken back to the project list.

## Steps to Reproduce

1. Go to project
2. Click on Deploy
3. Deploy any application
4. Click on the resources tab
5. Click on Deployments
6. Click on the recently deployed resource
7. Click delete

## Expected Behavior

I'd expect the resource pane to close, and the resource eventually removed from the list or marked for deleting (understanding k8s works on eventual consistency). I should also be kept in the Resources tab in the project.

## Actual Behavior

I am taken back to the project list and requires multiple clicks to get back.

## Screenshots

<img width="1495" height="1098" alt="Image" src="https://github.com/user-attachments/assets/8875e099-a8e4-46a8-9384-991540b9f9ac" />

## Environment

- **Plugin Version**: [e.g. 0.9.1]
- **Headlamp Version**: [e.g. 0.43.0]

## Logs

If applicable, please include relevant logs:

```
Paste logs here
```

## Checklist

- [ ] I have searched for existing issues
- [ ] I have provided all necessary information
- [ ] I have tested with the latest version
- [ ] I have included steps to reproduce

RounakKumarAgarwal:
I would like to work on this bug @pauldotyu , Could you assign it to me?

gambtho:
Root-caused, but holding it for a bit.

Two things going wrong:

1. `DeleteButton` hardcodes `startUrl: item.getListLink()`, which `ActionsNotifier`
   turns into a `history.push()` — so delete always bounces you to the cluster-wide
   resource list, outside the project. The project Resources tab doesn't override it.
2. Nothing clears the details drawer's `selectedResource` when the object is deleted,
   so the pane sits there on a resource that's gone.

Both are in Headlamp, not the plugin, and #823 is mid-flight reworking how we consume
Headlamp (submodule → patch series). Fixing it now means doing it twice, so we're
waiting on that to settle.

Neither bug is AKS-specific either — "delete shouldn't navigate you off the page you're
on" is just general Headlamp behaviour — so this probably wants to go upstream to
kubernetes-sigs/headlamp rather than living downstream.

pauldotyu:
> I would like to work on this bug [Paul Yu (@pauldotyu)](https://github.com/pauldotyu) , Could you assign it to me?

[...]

PR #890, opened 2026-08-25
upstreamable: don't navigate away or strand the details pane on delete

gambtho:
Fixes the two behaviours reported in #852: deleting a resource from inside a project threw the user back to the cluster-wide resource list, and left the details pane open on the deleted object.

Both are general Headlamp behaviour rather than anything AKS-specific, hence `upstreamable:` — this is a candidate for kubernetes-sigs/headlamp.

## Bug 1 — delete navigates you off the page you're on

`DeleteButton` hardcoded `startUrl`/`errorUrl` to `item.getListLink()`. `executeClusterAction` copies that into the action's `url`, and `ActionsNotifier` turns it into a `history.push()`. So delete always navigated to the cluster-wide list, no matter where it was triggered from.

Now it navigates away only when the page being viewed *is* the object's own details page — the one case that would 404 once the object is gone. Everywhere else (a list, a project view, the details pane) the user stays put. This is what `DeleteMultipleButton` already does with `startUrl: location.pathname`. An explicit `options` prop still wins, so callers and plugins keep control.

## Bug 2 — the details pane stays open

Worth noting for reviewers: the issue thread originally attributed this to `drawerMode.selectedResource` never being cleared. That state is actually **dead** — the only `dispatch(setSelectedResource(...))` in the tree is `DetailsDrawer.tsx:39` setting it to `undefined`, so `DetailsDrawer` never renders at all.

The pane is an **Activity**, launched by `Link.tsx` with `location: 'split-right'`, `temporary: true` — which is what `ProjectResourcesTab` renders resource names through. Nothing closed it on delete.

`MainInfoSectionHeader` is the one place that knows it is rendering the very object being deleted, so it now dismisses the enclosing activity on confirm via `useActivity()` + `Activity.close()`. Route-level details pages render outside any activity (`RouteSwitcher` and `ActivitiesRenderer` are siblings in `Layout`), so this is a no-op there, and row-level delete buttons are unaffected.

The pane closes immediately on confirm. Note this is inside the 5s undo grace period: cancelling restores the location but does not reopen the pane. That was a deliberate call over closing on success.

## Tests

New `DeleteButton.test.tsx` and `MainInfoSectionHeader.test.tsx`. The two behaviour-preservation cases (navigate away from a real details page; explicit `options` wins) were written to pass *before* the fix, so they guard against over-correction.

## Verification

- `npm test -- --run` — 107 files passed, 1 skipped; 1644 tests passed, 0 failed (includes 553 storyshots)
- `npm run ci-lint` (`--max-warnings 0`) — clean
- `npm run format-check` — clean
- `npx tsc --noEmit` — 4 errors, byte-identical to the unmodified branch tip; all pre-existing missing-dep resolution failures. `npm run tsc` could not be used: `tsgo` (`@typescript/native-preview`) is declared in `package.json` but missing from the installed tree.

Full suite verified at the branch tip; not re-run per-commit.

## Reviewer focus

The fix compares `location.pathname === item.getDetailsLink()`. If a multi-cluster details URL is ever formatted differently from `getDetailsLink()`'s output, the effect is that the user stays on a now-empty details page instead of being sent to the list — degraded rather than broken, but worth a second pair of eyes.

Does **not** bump the superproject submodule pointer, which is currently 9 commits behind `headlamp-downstream`. That's a separate PR.

### DIFF
--- a/frontend/src/components/common/Resource/DeleteButton.test.tsx
+++ b/frontend/src/components/common/Resource/DeleteButton.test.tsx
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { fireEvent, render, screen, within } from '@testing-library/react';
+import { KubeObject } from '../../../lib/k8s/KubeObject';
+import { TestContext } from '../../../test';
+import DeleteButton from './DeleteButton';
+
+const { mockClusterAction } = vi.hoisted(() => ({ mockClusterAction: vi.fn() }));
+
+vi.mock('../../../redux/clusterActionSlice', async importOriginal => ({
+  ...(await importOriginal<typeof import('../../../redux/clusterActionSlice')>()),
+  clusterAction: (...args: any[]) => {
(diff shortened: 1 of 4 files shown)


illume (approved):
🎉 thanks!

PR #915, opened 2026-09-01
fix Azure CLI bundling across platforms

ashu8912:
## Summary

Fix Azure CLI bundling and runtime discovery across Linux, macOS, and Windows.

- resolve and fingerprint the intended platform and app architecture during staging
- use native x64/ARM64 Linux workers and explicit packaging targets
- use the official x64 portable Azure CLI on Windows, including ARM64 emulation
- preserve the notarization-compatible x64 helper-tool policy on both macOS targets
- isolate bundled Windows extensions from the user profile
- reject stale or incomplete bundles and enforce executable/checksum verification
- update the Headlamp submodule to install bundled paths before environment caching

## Validation

Platforms exercised for this update:

- macOS ARM64 host: build/configuration test suite (30 tests passed)
- Linux x64: existing 1ES packaging path and target-specific workflow configuration
- Linux ARM64, macOS x64, and Windows x64/ARM64: architecture selection, checksums, staging markers, and wrapper behavior covered by unit tests; full installers were not run locally

Additional validation completed by CI:

- plugin packaging, plugin tests, lint/format, builds, translations, CodeQL, and security scans passed on the previous revision
- Electron command tests and TypeScript checks passed for the dependent Headlamp change

## Related issues and pull requests

Fixes #919

The dependent Headlamp changes are in #916.

PR #916, opened 2026-09-01
Fix bundled Azure CLI runtime selection

ashu8912:
## Summary

Fix bundled Azure CLI selection after restart, especially on Windows.

- install bundled-tool directories before backend, renderer, or MCP environment caching
- merge current bundled paths into command environments while preserving explicit per-command PATH overrides
- normalize Windows Path/PATH aliases and deduplicate entries case-insensitively
- add regression coverage for stale and overridden command environments
- keep macOS helper tools on the established notarization-compatible AMD64 policy; external-tools remains excluded from app signing

## Validation

- Electron TypeScript check passed
- electron/runCmd.test.ts: 41 passed

Parent packaging changes are in #915.

### aks-mcp hardening

The built-in server now resolves directly to `resources/external-tools/bin/aks-mcp.exe` on Windows (and `aks-mcp` elsewhere), while preserving custom commands. This removes PATH/PATHEXT dependency during fresh-install startup. Added Windows-specific executable resolution tests.

### DIFF
--- a/app/electron/plugin-management.ts
+++ b/app/electron/plugin-management.ts
@@ -1112,6 +1112,51 @@ export function getPluginBinDirectories(pluginsDir: string): string[] {
 }
 
 /**
+ * Removes empty and duplicate PATH entries while preserving order.
+ *
+ * addToPath can run more than once per session (bundled plugins, user plugins),
+ * so without this the same bin directories accumulate on every call. Windows
+ * paths are compared case-insensitively because the filesystem is too.
+ *
+ * @param entries - The candidate PATH entries, most significant first.
+ * @returns The deduplicated entries.
+ */
+function dedupePathEntries(entries: string[]): string[] {
+  const seen = new Set<string>();
+
+  return entries.filter(entry => {
+    if (!entry) return false;
+    const key = process.platform === 'win32' ? entry.toLowerCase() : entry;
+    if (seen.has(key)) return false;
+    seen.add(key);
+    return true;
+  });
+}
+
+/**
+ * Reads the current PATH and rewrites it under a single key.
+ *
+ * Windows environment variables are case-insensitive, but `process.env` is a
+ * plain object, so `Path` and `PATH` can coexist and spawned processes may read
+ * whichever one they find first. Deleting every casing before assigning keeps a
+ * single authoritative `PATH`.
+ *
+ * @param buildPath - Receives the existing PATH and returns the new one.
+ */
+function setProcessPath(buildPath: (existingPath: string) => string): void {
+  if (process.platform !== 'win32') {
+    process.env.PATH = buildPath(process.env.PATH || '');
+    return;
+  }
+
+  const existingPath = process.env.PATH || process.env.Path || '';
+  for (const key of Object.keys(process.env)) {
+    if (key.toLowerCase() === 'path') delete process.env[key];
+  }
+  process.env.PATH = buildPath(existingPath);
+}
+
+/**
  * Adds directories to the PATH environment variable
  *
  * @param dirs - Directories to add to PATH
@@ -1121,8 +1166,9 @@ export function addToPath(dirs: string[], description: string): void {
   if (dirs.length === 0) return;
 
   const pathSeparator = process.platform === 'win32' ? ';' : ':';
-  const existingPath = process.env.PATH || '';
-  process.env.PATH = [...dirs, existingPath].join(pathSeparator);
+  setProcessPath(existingPath =>
+    dedupePathEntries([...dirs, ...existingPath.split(pathSeparator)]).join(pathSeparator)
+  );
   const message = `Added ${dirs.length} ${description} bin directories to PATH`;
   console.info(message);
 }
--- a/app/electron/runCmd.ts
+++ b/app/electron/runCmd.ts
@@ -298,10 +298,7 @@ export async function executeCommandWithShellEnv(
       const child = spawn(command, args, {
         ...options,
         shell: useShell,
-        env: {
-          ...shellEnv,
-          ...options.env,
-        },
+        env: mergeCommandEnvironment(shellEnv, options.env),
       });
 
       let stdout = '';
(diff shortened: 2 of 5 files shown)


app/package.json
```diff
@@ -107,14 +105,7 @@
...
       "provisioningProfile": "mac/aks-desktop.provisionprofile",
-      "target": [
-        {
-          "target": "dmg",
-          "arch": [
-            "arm64"
-          ]
-        }
-      ],
+      "target": "dmg",
```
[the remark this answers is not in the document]
ashu8912: The PR description was stale. c9f5984/e4a50c7 deliberately restore the established notarization-compatible policy: macOS helper tools are AMD64 for both app targets and external-tools remains excluded from signing. The app itself is still packaged explicitly for x64 and ARM64.

app/package.json
```diff
@@ -107,20 +105,12 @@
...
```
skoeva, line 108: I think this change is related to e4a50c700c10e5dabe472793739fdd0de0f5377a ?

app/electron/plugin-management.ts
```diff
@@ -1121,8 +1121,22 @@ export function addToPath(dirs: string[], description: string): void {
   if (dirs.length === 0) return;
 
   const pathSeparator = process.platform === 'win32' ? ';' : ':';
-  const existingPath = process.env.PATH || '';
-  process.env.PATH = [...dirs, existingPath].join(pathSeparator);
+  const existingPath = process.env.PATH || process.env.Path || '';
```
illume: Please add a named function here, with comments explaining what this code is for?

app/electron/runCmd.ts
```diff
@@ -328,6 +325,27 @@ export async function executeCommandWithShellEnv(
...
+ * Merge a discovered shell environment with current process changes.
+ * Windows treats environment keys case-insensitively, while JavaScript does
+ * not; emitting both Path and PATH can make child_process choose the stale one.
+ */
+export function mergeCommandEnvironment(
+  shellEnv: NodeJS.ProcessEnv,
+  overrides: NodeJS.ProcessEnv = {}
+): NodeJS.ProcessEnv {
+  const merged = { ...shellEnv, ...overrides };
+  const currentPath = process.env.PATH ?? process.env.Path ?? merged.PATH ?? merged.Path;
```
illume: IMHO, we should only look at "Path" on windows, and keep all the windows logic inside the `if (process.platform === 'win32') {` guard.

app/package.json
illume: Not sure why this is done... can you please leave a note in the commit message body?

illume (commented):
I would prefer this was done upstream in the headlamp repo first.

app/package.json
```diff
@@ -73,15 +73,13 @@
...
             "x64",
-            "armv7l",
             "arm64"
           ]
         },
         {
           "target": "tar.gz",
           "arch": [
             "x64",
-            "armv7l",
```
illume, line 84: Why is this removed?

app/package.json
```diff
@@ -73,15 +73,13 @@
           "target": "AppImage",
           "arch": [
             "x64",
-            "armv7l",
```
illume, line 76: Why is this removed?

illume (changes requested):
Thanks for this.

Please see open review comments? I think this needs cleaning up a bit.

The git messages need to follow commit guidelines. It seems the commits are not atomic (see cp review message), so please check that?

I would prefer this was done upstream in the headlamp repo first.

app/package.json
ashu8912: Yes — and both are now explained in the commit body of `b22c1561f` (squashed). Short version: the previous `[{ "target": "dmg", "arch": ["arm64"] }]` form pins the architecture in config, which overrides the `--x64` / `--arm64` flag that `build/build-mac-target.ts` passes to electron-builder, so an x64 build silently produced an arm64 dmg. `signIgnore` keeping `external-tools` is unchanged from the base branch — that was a mistaken removal that I have folded away rather than leaving as a follow-up commit.

app/electron/plugin-management.ts
ashu8912: Done. Extracted `dedupePathEntries()` and `setProcessPath()`, both with doc comments explaining why the dedupe and the Windows `Path`/`PATH` collapse are needed. `addToPath()` is now three lines.

app/electron/runCmd.ts
ashu8912: Done. `mergeCommandEnvironment()` now returns early from inside the `if (process.platform === 'win32')` branch, and `Path` is only consulted there. The non-Windows path is just `overrides.PATH ?? process.env.PATH ?? shellEnv.PATH`.

app/electron/runCmd.ts
```diff
@@ -403,10 +423,10 @@ export async function handleRunCommand(
...
     ...commandData.options,
     shell: useShell,
-    env: {
-      ...shellEnv,
-      ...(commandData.command === 'scriptjs' ? { HEADLAMP_RUN_SCRIPT: 'true' } : {}),
-    },
+    env: mergeCommandEnvironment(
+      shellEnv,
+      commandData.command === 'scriptjs' ? { HEADLAMP_RUN_SCRIPT: 'true' } : {}
+    ),
```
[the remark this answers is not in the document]
ashu8912: Fixed — `commandData.options.env` is now spread into the overrides, with the `scriptjs` flag layered on top. Added a regression test (`handleRunCommand - command environment`) asserting a per-command env var reaches the spawned process.

app/electron/runCmd.ts
```diff
@@ -328,6 +325,29 @@ export async function executeCommandWithShellEnv(
...
 
+/**
+ * Merge a discovered shell environment with current process changes.
+ * Windows treats environment keys case-insensitively, while JavaScript does
+ * not; emitting both Path and PATH can make child_process choose the stale one.
+ */
+export function mergeCommandEnvironment(
+  shellEnv: NodeJS.ProcessEnv,
+  overrides: NodeJS.ProcessEnv = {}
+): NodeJS.ProcessEnv {
```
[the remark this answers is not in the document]
ashu8912: Done — squashed `d077c94`, `e4a50c7` and `f04e358` into a single `aksd: fix bundled Azure CLI runtime selection` commit, with a body covering both the runtime and packaging rationale. The branch is now two commits.

app/package.json
ashu8912: Explained in the squashed commit body: there are no Azure CLI or aks-mcp release assets for `armv7l` — `build/az-cli-config.ts` (`SUPPORTED_ARCHES = {x64, arm64}`) and `build/download-aks-mcp.ts` both reject it — so an armv7l package would ship without the bundled tools this PR depends on.

app/package.json
ashu8912: Same reason as the AppImage target: no `armv7l` release assets exist for the bundled Azure CLI or aks-mcp, so that package would ship without its bundled tools. Rationale is now in the commit body.

illume:
I'll take over this one.

Issue #891, opened 2026-08-25
[BUG] Project Access Returns No Results


willspacecase:
## Bug Description

When creating an AKS Project, the user search field fails to return results when searching for users to assign project access. Instead of displaying a searchable dropdown of Microsoft Entra ID users, the field remains empty and no users can be selected.

## Steps to Reproduce

1. Open AKS desktop.
2. Navigate to Projects.
3. Select **Create Project**.
4. Choose an eligible AKS cluster.
5. Proceed to the **Project Access** step.
6. Begin typing a user's name or email address in the user search field.
7. Observe that no search results are returned.

## Expected Behavior

The Project Access user picker should display matching Microsoft Entra ID users as the user types, allowing project administrators to easily search for and select users to grant project access.

## Actual Behavior

No search results are displayed when typing in the Project Access user search field. Users cannot be selected through the search experience, preventing project administrators from assigning access using the expected workflow.

## Screenshots

<img width="3006" height="1854" alt="Image" src="https://github.com/user-attachments/assets/ba88af68-7843-433d-927e-c3e1563db5ac" />

## Environment

- **Plugin Version - AKS desktop**: [0.9.0]
- **Headlamp Version**: [0.43.0]
- **Kubernetes Version**: [1.35.6]
- **Azure CLI Version**: [na]
- **Browser**: [na]
- **OS**: [e.g. macOS 26.6.1]

## Additional Context

The issue was observed while testing AKS Projects. The expected experience is a searchable dropdown that returns users from Microsoft Entra ID. The behavior appears inconsistent across environments and may be related to the user lookup functionality rather than project creation itself.

## Logs

If applicable, please include relevant logs:

```
Paste logs here
```

## Checklist

- [ ] I have searched for existing issues
- [ ] I have provided all necessary information
- [ ] I have tested with the latest version
- [ ] I have included steps to reproduce

skoeva:
I reproduced this on macOS with AKS Desktop 0.9.0. The underlying `az ad user list` call fails with `AADSTS530084` because Conditional Access token protection blocks Microsoft Graph access.

I plan to update the picker to distinguish lookup failures from “No users found,” show an actionable error, and support manual Entra Object ID entry. Checking the reporter’s app logs would confirm whether this is the same root cause.

skoeva:
search does not work under our azure subscription

potential follow-up work: Add message for new users to find object id

PR #974, opened 2026-09-15
aks-desktop: Add fallback for blocked Entra user search

skoeva:
AKS desktop treated Entra directory lookup failures as empty results, leaving users without useful feedback when Conditional Access or permissions blocked Microsoft Graph. This distinguishes lookup failures from genuine empty results and provides a reliable manual Object ID or fallback.

Related: #891 

### Changes

- Show distinct states for searching, no matches, lookup failures, and unavailable directory search
- Detect ConditionalAccess and permission errors for both user search and identifier resolution
- Accept manually entered Object IDs without requiring directory resolution
- Clear stale accepted IDs when the input is edited
- Add localization keys for the now user-facing messages

### Testing

- [x] 36 focused unit tests pass
- [x] `npm run tsc` passes
- [x] `npm run lint` passes
- [x] Manually verified Conditional Access fallback and Object ID any on MacOS

#### Valid object ID entered

image

#### Unknown user credential entered

image

#### No credential entered

image


plugins/aks-desktop/src/utils/azure/az-ad.ts
```diff
@@ -15,6 +15,16 @@ export interface AzureADUser {
   userPrincipalName: string;
 }
 
+export function isAzureADLookupUnavailable(error?: string): boolean {
```
skoeva, line 19: here are the relevant messages:

- `AADSTS530084: Access has been blocked by conditional access token protection policy configured by this organization. To learn more, see https://aka.ms/TBCADocs.`
- `AADSTS50079: Due to a configuration change made by your administrator, or because you moved to a new location, you must enroll in multifactor authentication to access this resource.`

illume: Can you please add docs for this function with something like that in it?

illume (dismissed):
🎉 thanks!

illume (approved):

PR #987, opened 2026-09-23
build: Allow plugin catalog to proxy Artifact Hub API

skoeva:
AKS desktop's packaged backend rejected plugin catalog requests because Artifact Hub was missing from the proxy allowlist, leaving the catalog loading indefinitely. This permits only the required package API.

### Changes

- Allow proxying `https://artifacthub.io/api/v1/packages/*`
- Add regression coverage for the product configuration

### Testing

- [x] Product manifest tests pass
- [x] Catalog search and details requests return HTTP 200
- [x] Unrelated Artifact Hub paths remain denied


illume (approved):
🎉 thanks!

